PR #20, opened 2022-02-19
Angular Kinematics

moorepants:
https://htmlpreview.github.io/?https://github.com/moorepants/learn-multibody-dynamics/blob/angular-kinematics-build/index.html

### DIFF
--- a/angular.rst
+++ b/angular.rst
@@ -505,8 +505,8 @@ and with manual derivatives:
 
    D.ang_vel_in(A).dt(D).simplify()
 
-Addition of Angular Velocity
-============================
+Addition of Angular Acceleration
+================================
 
 The calculation of angular acceleration is relatively simple, but the addition
 of angular velocities explained in Sec. :ref:`addition-of-angular-velocity`
--- a/angular.rst
+++ b/angular.rst
@@ -10,12 +10,13 @@ Angular Kinematics
 Angular Velocity
 ================
 
-If the relative orientation of two reference frames change with respect to
-time, then we can calculate the angular velocity :math:`{}^A\bar{\omega}^B` of
-reference frame :math:`B` in reference frame :math:`A`. If
-:math:`\hat{b}_x,\hat{b}_y,\hat{b}_z` are right handed mutually perpendicular
-unit vectors fixed in :math:`B` then the angular velocity of :math:`B` in
-:math:`A` is defined as ([Kane1985]_, pg. 16):
+In Ch. :ref:`orientation` we learned that reference frames can be oriented
+relative to each other. If the relative orientation of two reference frames
+change with respect to time, then we can calculate the angular velocity
+:math:`{}^A\bar{\omega}^B` of reference frame :math:`B` in reference frame
+:math:`A`. If :math:`\hat{b}_x,\hat{b}_y,\hat{b}_z` are right handed mutually
+perpendicular unit vectors fixed in :math:`B` then the angular velocity of
+:math:`B` in :math:`A` is defined as ([Kane1985]_, pg. 16):
 
 .. math::
 
@@ -26,7 +27,8 @@ unit vectors fixed in :math:`B` then the angular velocity of :math:`B` in
 
 If :math:`B` is oriented with respect to :math:`A` and mutually perpendicular
 unit vectors :math:`\hat{a}_x,\hat{a}_y,\hat{a}_z` are fixed in :math:`A` then
-there are these general relationships among the unit vectors of each frame:
+there are these general relationships among the unit vectors of each frame (see
+:ref:`direction-cosine-matrix`):
 
 .. math::
 
@@ -117,8 +119,8 @@ The angular velocity vector is then:
    A_w_B = mnx*B.x + mny*B.y + mnz*B.z
    A_w_B
 
-Simple Rotations
-================
+Simple Orientations
+===================
 
 For a simple orientation about the :math:`z` axis through :math:`\theta` the
 direction cosine matrix is:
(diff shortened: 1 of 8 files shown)


moorepants:
@jtheinen The draft for 4.1 Angular Kinematics is done here. I haven't merged it so the rendered version doesn't show up on the website yet, but if I leave it as a pull request then you can comment inline and even make suggestions for changes in the Github interface. So either you can 1) review the raw rst text, 2) review the raw rst text and build the website on your own computer to see the rendered version or 3) I can merge now and you can see it on the live website and make your own review issue. What would you like?

moorepants:
I currently don't have any nice examples of real systems in this chapter. I should add one, but if I don't have time this week I'll just do the example in the lecture video.

moorepants:
If you want to go with option 2) I added build instructions here: https://github.com/moorepants/learn-multibody-dynamics#building-the-website

jtheinen:
< @jtheinen The draft for 4.1 Angular Kinematics is done here. I haven't merged ......>

I think it is better if I check it in website version, maybe I'll spot some lay-out corrections. So 1) is not preferred. Let's go with 2) since students won't be able to see the draft then. 

I made the html files but I don't seem to get the angular kinematics file when following the build instructions

I'll work on it every Tuesday afternoon, will that work for you?

jtheinen:
I will do inline comments, just show me in our meeting how to include the pull request when I make the website

angular.rst
```diff
@@ -0,0 +1,567 @@
...
+
+.. warning::
+
+   This page is a draft until February 25, 2022. Report issues at:
+   https://github.com/moorepants/learn-multibody-dynamics/issues
+
+Angular Velocity
+================
+
+If the relative orientation of two reference frames change with respect to
```
jtheinen, line 13: Refer to previous chapter: 

We seen in section orientations that frames can be oriented relative to each other.

angular.rst
```diff
@@ -0,0 +1,567 @@
...
+   \left(\frac{{}^A d\hat{b}_z}{dt} \cdot \hat{b}_x\right) \hat{b}_y +
+   \left(\frac{{}^A d\hat{b}_x}{dt} \cdot \hat{b}_y\right) \hat{b}_z
+
+If :math:`B` is oriented with respect to :math:`A` and mutually perpendicular
+unit vectors :math:`\hat{a}_x,\hat{a}_y,\hat{a}_z` are fixed in :math:`A` then
+there are these general relationships among the unit vectors of each frame:
+
+.. math::
+
+  \hat{b}_x & = c_{xx} \hat{a}_x + c_{xy} \hat{a}_y + c_{xz} \hat{a}_z \\
```
jtheinen, line 33: This is the same principle as the direction cosine matrix right?

moorepants: Yes, same way I defined it in the orientation chapter.

angular.rst
```diff
@@ -0,0 +1,567 @@
...
+
+.. jupyter-execute::
+
+   A_w_B = mnx*B.x + mny*B.y + mnz*B.z
+   A_w_B
+
+Simple Rotations
+================
+
+For a simple orientation about the :math:`z` axis through :math:`\theta` the
```
jtheinen, line 123: orientation = rotation

angular.rst
```diff
@@ -0,0 +1,567 @@
...
+
+   B_C_A = sm.Matrix([[sm.cos(theta), sm.sin(theta), 0],
+                      [-sm.sin(theta), sm.cos(theta), 0],
+                      [0, 0, 1]])
+   B_C_A
+
+Following the same pattern as before the angular velocity of :math:`B` in
+:math:`A` can be formed:
+
+.. note::
```
jtheinen, line 138: Maybe move note to line 176, a bit distracting here

angular.rst
moorepants: I've been back and forth on this a bit. The only time I've said rotation, I think, is about the simple rotation. But I use orientation otherwise. So I'm thinking to use orientation always to be consistent. Let me think about that.

angular.rst
```diff
@@ -0,0 +1,567 @@
...
+.. todo:: Should this return the angular velocity expressed in the body fixed
+   frame?
+
+A simple rotation and associated simple angular velocity can be formulated for
+any arbitrary rotation axis vector, not just one of the three mutually
+perpendicular unit vectors as shown above. There is a simple angular velocity
+between two reference frames :math:`A` and :math:`B` if there exists a single
+unit vector :math:`\hat{k}` which is fixed in both :math:`A` and :math:`B` for
+some finite time. If this is the case, then :math:`{}^A\bar{\omega}^B = \omega
+\hat{k}` where :math:`\omega` is the time rate of change of the angle
```
jtheinen, line 183: is time rate of change same as rate of change?

angular.rst
```diff
@@ -0,0 +1,567 @@
...
+.. note:: :math:`\sqrt{x^2}=x` is only true if :math:`x` is real and
+   :math:`x>0`.
+
+.. todo:: Why doesn't this simplify to theta dot? I tried ``real=True`` on
+   theta.
+
+Body Fixed Rotations
+====================
+
+Similarly, if you establish a Euler :math:`z\textrm{-}x\textrm{-}z` orientation
```
jtheinen, line 217: Why similarly? because A and B in previous example share a fixed unit vector?

angular.rst
```diff
@@ -0,0 +1,567 @@
...
+   A_w_B = B.ang_vel_in(A)
+   A_w_B
+
+:math:`\frac{{}^Ad\bar{u}}{dt}`: is then:
+
+.. jupyter-execute::
+
+   u.dt(B) + me.cross(A_w_B, u)
+
+We can show that first results is equivalent by expressing in :math:`B` and
```
jtheinen, line 347: add 'the' 
we can show that 'the' first ...

angular.rst
```diff
@@ -0,0 +1,567 @@
...
+
+   u.express(A).dt(A).express(B).simplify()
+
+.. _addition-of-angular-velocity:
+
+Addition of Angular Velocity
+============================
+
+Similar to the relationship in direction cosine matrices of successive
+orientations (Sec. :ref:`successive-orientations`), there is a relationship
```
jtheinen, line 360: state that it is now addition in stead of dot product (Maybe this is directly obvious from eq, then its unnecessary to add)

angular.rst
```diff
@@ -0,0 +1,567 @@
...
+   C_alp_D
+
+Summing the successive angular accelerations gives this result:
+
+.. jupyter-execute::
+
+   A_alp_D = A_alp_B + B_alp_C + C_alp_D
+   A_alp_D.express(D).simplify()
+
+which is not equal to the correct, more complex, result above!
```
jtheinen, line 565: Unclear what is not equal to what and which is correct.

moorepants:
I addressed all the comments here. Thanks! The page is now live.

PR #25, opened 2022-02-20
Translational kinematics

moorepants:


moorepants:
@jtheinen this is ready for review too. It'd a bit rougher, but I can polish before Friday. It's complete for review purposes.

moorepants:
Also, in this one, ignore the files from #20 because those should be reviewed in that PR.

translational.rst
```diff
@@ -0,0 +1,547 @@
...
+
+where point :math:`T` is a point that conicides with :math:`R` at that instant.
+
+Combined with the two point theorem for :math:`T`, you can write:
+
+.. math::
+
+   {}^N\bar{v}^R = {}^B\bar{v}^R + {}^N\bar{v}^S + {}^N\bar{\omega}^B \times \bar{r}^{R/S}
+
+In our kinetic sculpture example, ff the pigeon :math:`R` is walking at a
```
jtheinen, line 319: ff = if

jtheinen (commented):
The chapter translational kinematics is really understandable, coherent and well explained!

jtheinen:
I was thinking, maybe it would be more understandable to start with translational kinematics then do angular kinematics. It is easier to comprehend the translational chapter and you could even use the same example of the kinetic structure in rotterdam for the angular velocity, body fixed rotations and addition of angular acceleration

moorepants:
> I was thinking, maybe it would be more understandable to start with translational kinematics then do angular kinematics. It is easier to comprehend the translational chapter and you could even use the same example of the kinetic structure in rotterdam for the angular velocity, body fixed rotations and addition of angular acceleration

I think I need to keep them in that order, but I could introduce the example in the angular chapter and use it in both. Good idea.

moorepants:
@jtheinen you only found a single error in this page?

PR #33, opened 2022-02-28
Configuration and generalized coordinate chapter.

moorepants:


moorepants:
@jtheinen this is ready. This week it is just the translational one and this one. You read the translational one. I've found a handful of math errors in the prior pages. Try to read through the math as carefully as you can, because I'm way more likely to make mistakes there.

configuration.rst
```diff
@@ -0,0 +1,299 @@
...
+
+Consider the linkage shown below:
+
+.. _configuration-four-bar:
+.. figure:: figures/configuration-four-bar.svg
+   :align: center
+   :width: 600px
+
+   a) Shows for links :math:`A`, :math:`B`, :math:`C`, and :math:`N` with
+   respective lengths :math:`l_a,l_b,l_c,l_n` connected at points
```
jtheinen, line 16: maybe say connected with 2D joints

configuration.rst
```diff
@@ -0,0 +1,299 @@
...
+.. jupyter-execute::
+
+   fh = sm.Matrix([fhx, fhy])
+   fh
+
+General Holonomic Constraints
+=============================
+
+If you consider the points :math:`P_1,P_2,\ldots,P_v` as unconstrained in
+Euclidean space, then we would need :math:`3v` constraint equations to fully
```
jtheinen, line 156: Unclear what you mean by this

configuration.rst
```diff
@@ -0,0 +1,299 @@
...
+   :label: p4-constraint
+
+   \bar{r}^{P_4/P_1}\cdot \hat{n}_y = 0
+
+These 11 constraints leave a single free coordinate to describe the orientation
+of :math:`A`, :math:`B`, and :math:`C` in :math:`N`. When we originally
+sketched :numref:`configuration-four-bar` most of these constraints were
+implied, i.e. we drew a planar mechanism with points :math:`P_1` and
+:math:`P_4` fixed in :math:`N`, but formally there are 12 coordinates needed to
+locate the four points and 11 constraints that constrain them to have the
```
jtheinen, line 211: I count only 10 constraints in the given example

configuration.rst
```diff
@@ -0,0 +1,299 @@
...
+configuration of a four-bar linkage.
+
+A general holonomic constraint for a set of :math:`v` points with Cartesian
+coordinates is then ([Kane1985]_ pg. 35):
+
+.. math::
+   :label: holonomic-cartesian
+
+   f_h(x_1, y_1, z_1, \ldots, x_v, y_v, z_v, t) = 0
```
jtheinen, line 220: The t term is not introduced yet

configuration.rst
```diff
@@ -0,0 +1,299 @@
...
+
+Generalized Coordinates
+=======================
+
+If a set of :math:`v` points are constrained with :math:`M` holonomic
+constraints then only :math:`n` of the Cartesian coordinates are independent
+of each other. The number of independent coordinates is found with
```
jtheinen, line 227: Would it make sense to mention degrees of freedom, students know this from their bachelor

configuration.rst
```diff
@@ -0,0 +1,299 @@
...
+([Kane1985]_ pg 37):
+
+.. math::
+   :label: num-gen-coord
+
+   n := 3v - M
+
+These :math:`n` independent Cartesian coordinates can also be expressed as
+:math:`n` functions of time :math:`q_1(t),q_2(t),\ldots,q_n(t)` in such a way
```
jtheinen, line 236: Is a generalized coordinate only possible as a function of time? you can also describe a systems configuration with a generalized coordinate without time right?

configuration.rst
```diff
@@ -0,0 +1,299 @@
...
+
+.. figure:: figures/configuration-pendulum.svg
+   :align: center
+   :width: 400px
+
+If the pendulum length :math:`l` is constant and the orientation between
+:math:`A` and :math:`N` can change, then the location of :math:`P` relative to
+:math:`O` can be described with the Cartesian coordinates :math:`x` and
+:math:`y`. It should be clear that :math:`x` and :math:`y` depend on each other
+for this system. The constraint relationship is between those two coordinates
```
jtheinen, line 253: delete first 'is'

configuration.rst
```diff
@@ -0,0 +1,299 @@
...
+:math:numref:`pendulum-length-constraint`. For example, these two coordinates
+can also be written as as functions of the angle :math:`q`:
+
+.. math::
+   :label: xy-func-of-q
+
+   x = \cos q \\
+   y = \sin q
+
+and if we describe the configuration with only :math:`q`, the constraint is
```
jtheinen, line 277: Maybe state here why it is helpful to describe a system in one coordinate. Now everybody sees how the generalized coordinate is formed, but why would you do it anyway. Simplifying systems, less computational power, not having to write down all constraints

configuration.rst
```diff
@@ -0,0 +1,299 @@
...
+Now, let's return to the four-bar linkage example in
+:numref:`configuration-four-bar` and think about what are the generalized
+coordinates of this system.
+
+.. admonition:: Exercise
+
+   Are :math:`q_1,q_2,q_3` generalized coordinates of the four-bar linkage? If
+   not, why?
+
+.. admonition:: Solution
```
jtheinen, line 290: The layout of solution is very narrow, does not cover all width of screen.
Schermafbeelding 2022-03-04 om 01 07 05

configuration.rst
```diff
@@ -0,0 +1,299 @@
...
+   :class: dropdown
+
+   Any one of the :math:`q_1,q_2,q_3` can be a generalized coordinate, but only
+   one. The other two are depdendent due to the two constraints. We started
+   with three coordinates :math:`q1,q_2,q_3` describing the open chain
+   :math:`P_1` to :math:`P_2` to :math:`P_3` to :math:`P_4`. Then we have two
+   scalar constraint equations, leaving :math:`n=1`. Thus we can choose
+   :math:`q_1`, :math:`q_3`, *or* :math:`q_3` to be the indepdendent
```
jtheinen, line 298: change q3 to q2

configuration.rst
moorepants: 4 points = 3 degrees of freedom * 4 points = 12 degree of freedom

4 points * 1 z constraint = 4 constraints
1 constraint per bar connecting points = 4 constraints
x and y constraint on point 1 = 2 constraints
y constraint on point 4 = 1 constraint

11 constraints, leaving 1 degree of freedom, no?

configuration.rst
moorepants: No, they are strictly about functions of time because if they are, the system can't move. It would just be a static configuration.

configuration.rst
moorepants: I think that is an issue with the htmlpreview and won't look like that in the regular site.

configuration.rst
moorepants: I think I want to avoid the word "joints" at the point. It would force me to define a joint and that potentially opens a can of worms. An appendix on joints could be nice though.

configuration.rst
moorepants: I'll write the one vector equation (that represents two constraints) in two separate dot products to clear it up.

configuration.rst
moorepants: Yes, I need to explain why that it there. Thanks.

configuration.rst
moorepants: I'm reluctant to introduce the term here. I plan to introduce in in the next chapter for motion constraints for sure, because: `dof = 3v - M - m` where m= number of motion constraints. Let me think about whether I can introduce it here.

configuration.rst
moorepants: Yes, I think I can expand that a bit and talk about how we only need 2 constraint equations in the four bar linkage if we use generalized coordinates instead of 11 if we don't.

configuration.rst
jtheinen: True, i just counted the equations but the fixation in 0 is obviously 2 constraints

configuration.rst
jtheinen: I could not do make html in the command line with this branch, probably because of the new layout right

moorepants:
@jtheinen I think I've addressed things sufficiently. Going to merge. Thanks!

PR #66, opened 2022-04-20
Fix typos; Reword some parts of the text

namannimmo10:
Ran the text through Grammarly. Hope this helps, @moorepants.

### DIFF
--- a/introduction.rst
+++ b/introduction.rst
@@ -16,7 +16,7 @@ Prerequisites
 
 - Linear algebra
 - Vector calculus
-- Calculus based physics
+- Calculus-based physics
 - Statics
 - Dynamics
 - Introductory numerical methods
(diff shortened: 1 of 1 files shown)


moorepants:
Naman, here are the things I took from your PR: 8210783

namannimmo10:
Thanks. Grammarly could still be useful if you don't want to hunt for typos manually. I guess I probably selected the formal text option, that's why it made some pointless changes.

moorepants:
I'll probably be opinionated about this because the text is written by me and it should feel like me, errors and all. Something like grammarly can be a bit over zealous in what it recommends when you personally may be ok with some grammar variations. The main things I'd like feedback on are clarity of concepts. If my grammar decreases the clarity of the ideas, then it should be fixed.

namannimmo10:
That makes sense!

PR #129, opened 2023-02-15
2023 updates to orientation chapter.

moorepants:
Updates:

- Added some new short exercises.
- Turned the final Euler angle explanation into an exercise.
- Adjust/improved wording.
- Added some more external links.
- Simplified the intersphinx links to sympy.
- [x] Add handwritten direction cosine matrix notation and unit vector notation.

### DIFF
--- a/orientation.rst
+++ b/orientation.rst
@@ -13,6 +13,18 @@ Orientation of Reference Frames
    import sympy as sm
    sm.init_printing(use_latex='mathjax')
 
+Learning Objectives
+===================
+
+After completing this chapter readers will be able to:
+
+- Define a reference frame with associated unit vectors.
+- Define a direction cosine matrix between two oriented reference frames.
+- Derive direction cosine matrices for simple rotations.
+- Derive direction cosine matrices for successive rotations.
+- Manage orientation and direction cosine matrices with SymPy.
+- Rotate reference frames using Euler Angles.
+
 Reference Frames
 ================
 
--- a/orientation.rst
+++ b/orientation.rst
@@ -2,8 +2,6 @@
 Orientation of Reference Frames
 ===============================
 
-.. warning:: This page as not yet been updated for the 2022-2023 course.
-
 .. note::
 
    You can download this example as a Python script:
@@ -196,23 +194,39 @@ SymPy can also find the transpose of this matrix;
 
    A_C_N.transpose()
 
-Notably, the inverse and the transpose are the same here. It turns out that
-this will be generally true for these matrices that describe the orientation
-between reference frames. Following the notation convention, this holds:
+Notably, the inverse and the transpose are the same here. This indicates that
+this matrix is an `orthogonal matrix`_. All matrices that describe the
+orientation between reference frames are orthogonal matrices. Following the
+notation convention, this holds:
 
 .. math::
    :label: dcm-inverse-transpose
 
    {}^N\mathbf{C}^A = \left({}^A\mathbf{C}^N\right)^{-1} = \left({}^A\mathbf{C}^N\right)^T
 
+.. _orthogonal matrix: https://en.wikipedia.org/wiki/Orthogonal_matrix
+
 .. admonition:: Exercise
 
-   Write :math:`{}^N\mathbf{C}^A` for a simple rotation about the shared
-   :math:`\hat{n}_y` and :math:`\hat{a}_y` axes, rotating :math:`A` with
-   respect to :math:`N`.
+   Write :math:`{}^N\mathbf{C}^A` for simple rotations about both the shared
+   :math:`\hat{n}_x` and :math:`\hat{a}_x` and shared :math:`\hat{n}_y` and
+   :math:`\hat{a}_y` axes, rotating :math:`A` with respect to :math:`N` through
+   angle :math:`\theta`.
 
 .. admonition:: Solution
    :class: dropdown
+
+   For a :math:`x` orientation:
+
+   .. math::
+
+      \begin{bmatrix}
+        1 &  0  & 0 \\
+        0 & \cos{\theta} & \sin{\theta} \\
+        0 & -\sin{\theta} & \cos{\theta}
+      \end{bmatrix}
+
+   For a :math:`y` orientation:
 
    .. math::
 
@@ -222,8 +236,8 @@ between reference frames. Following the notation convention, this holds:
         \sin{\theta} & 0 & \cos{\theta}
       \end{bmatrix}
 
-Direction Cosine Matrix
-=======================
+Direction Cosine Matrices
+=========================
 
 If now :math:`A` is oriented relative to :math:`N` and the pairwise angles
 between each :math:`\hat{a}` and :math:`\hat{n}` mutually perpendicular unit
@@ -302,17 +316,16 @@ prior section for a simple orientation is a specific case of this more general
 definition. The direction cosine matrix is also referred to as a "rotation
 matrix" or "orientation matrix" in some texts.
 
-.. todo:: Apply the general definition of the direction cosine matrix to a
-   simple rotation about x to show it is the same as we worked out above.
+.. todo:: Create a simple exercise involving applying the definition.
 
 Successive Orientations
 =======================
 
 Successive orientations of a series of reference frames provides a convenient
-way to describe an arbitrary orientation between a pair of reference frames.
-Below, an additional reference frame :math:`B` is shown that is simply oriented
-with respect to :math:`A` in the same way that :math:`A` is from :math:`N`
-above.
+way to manage orientation among more than a single pair. Below, an additional
+auxiliary reference frame :math:`B` is shown that is simply oriented with
+respect to :math:`A` in the same way that :math:`A` is from :math:`N` above in
+the prior section.
 
 .. _orientation-simple-successive:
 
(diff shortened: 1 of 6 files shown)


Peter230655:
_Vectors have a magnitude, direction, and sense (:math:`\pm`) but notably not a
 position_.
1.
I find the addition: _notably not a position_ to be very important! Maybe you could add something like: two vectors which differe only by a parallel displacement are (represented by) the same vector?
2.
I am not very clear, what _sense_ means. Does _direction_ not imply that the vector defines a line in space and which way to travel in the positive direction? ( May be may lack of command of English)

moorepants:
>  I am not very clear, what sense means. Does direction not imply that the vector defines a line in space and which way to travel in the positive direction? ( May be may lack of command of English)

I use "direction" to tell us what line in space the vector is parallel to and "sense" the positive or negative sense along that direction. I show it a bit more in the vectors chapter.

moorepants:
@wwolfie this is ready to be checked if you have time. Sorry that I'm late getting this updated. I'll try to do it earlier in the week next week.

moorepants:
> two vectors which differ only by a parallel displacement are (represented by) the same vector

This wording seems to imply that vectors can have a displacement. So I don't think I'll add this.

Maybe "even if vectors are drawn with an apparent parallel displacement they are the same". I show this in the figure though, maybe it just isn't as explicit as it could be.

Peter230655:
I admit I did not look at the picture. Your wording -if the picture does not say it- of course is much better.
I just find this _no position_ statement very important: An engineer thinks about a vector as being a force attacing at some point - at least this happened with me until I got better in math.

_Sense:_ clear now, thanks!

orientation.rst
```diff
@@ -83,7 +88,8 @@ of unit vectors are initially aligned, the :math:`A` frame can then be simply
 oriented about the common parallel :math:`z` unit vectors of the two frames. We
 then say "reference frame :math:`A` is oriented with respect to reference frame
 :math:`N` about the shared :math:`z` unit vectors through an angle
-:math:`\theta`. A visual representation of this orientation is:
+:math:`\theta`. A visual representation of this orientation looking into the
+:math:`z` unit vectors is:
```
wwolfie, line 92:
```suggestion
:math:`\theta`. A visual representation of this orientation looking from the direction of the positive :math:`z` unit vector is:
```

In any case, it should be vector (singular).

orientation.rst
```diff
@@ -188,23 +194,55 @@ SymPy can also find the transpose of this matrix;
 
    A_C_N.transpose()
 
-Notably, the inverse and the transpose are the same here. It turns out that
-this will be generally true for these matrices that describe the orientation
-between reference frames. Following the notation convention, this holds:
+Notably, the inverse and the transpose are the same here. This indicates that
+this matrix is an `orthogonal matrix`_. All matrices that describe the
```
wwolfie, line 198: special orthogonal matrix would be complete

```suggestion
this matrix is a `special orthogonal matrix`_. All matrices that describe the
```

orientation.rst
```diff
@@ -188,23 +194,55 @@ SymPy can also find the transpose of this matrix;
...
+.. _orthogonal matrix: https://en.wikipedia.org/wiki/Orthogonal_matrix
 
-Direction Cosine Matrix
-=======================
+.. admonition:: Exercise
+
+   Write :math:`{}^N\mathbf{C}^A` for simple rotations about both the shared
+   :math:`\hat{n}_x` and :math:`\hat{a}_x` and shared :math:`\hat{n}_y` and
+   :math:`\hat{a}_y` axes, rotating :math:`A` with respect to :math:`N` through
+   angle :math:`\theta`.
```
wwolfie, line 214: The direction of the angle is never specified. Maybe the default direction is best explained at the spot where also the "right-handedness" of the frames is specified?

orientation.rst
```diff
@@ -188,23 +194,55 @@ SymPy can also find the transpose of this matrix;
...
+.. admonition:: Solution
+   :class: dropdown
+
+   For a :math:`x` orientation:
+
+   .. math::
+
+      \begin{bmatrix}
+        1 &  0  & 0 \\
+        0 & \cos{\theta} & \sin{\theta} \\
```
wwolfie, line 225: It's early morning, but I have convinced myself that this is the inverse of the correct solution

orientation.rst
```diff
@@ -215,7 +253,9 @@ defined.  For example, the figure below shows the three angles
 
    Three angles relating :math:`\hat{a}_x` to the unit vectors of :math:`N`.
 
-Similarly to the simple example above, we can write these equations:
+Similar to the simple example above, we can write these equations if the
+:math:`\alpha_y` and :math:`\alpha_z` angles relate the :math:`\hat{a}_y` and
+:math:`\hat{a}_z` unit vectors to those of :math:`N`:
```
wwolfie, line 258: Easiest to add this comment here rather than make new PR:
In the sentence "Since we are working with mutually perpendicular unit vectors ...", 'mutually perpendicular' and the second 'unit' are not needed, which might confuse readers who do not know the dot product <-> cosine rule by heart.

orientation.rst
```diff
@@ -271,32 +307,36 @@ where the general direction cosine matrix is then:
...
-:math:`N` and :math:`A`, it is invertible, and its inverse is equal to the
-transpose, as shown above in the simple example. The direction cosine matrix
-found in the prior section for a simple orientation is a specific case of this
-more general definition. The direction cosine matrix is also referred to as a
-rotation matrix in some texts.
+We call :math:`{}^A\mathbf{C}^N` the "direction cosine matrix" as a general
+description of the relative orientation of two reference frames. This matrix
+uniquely defines the relative orientation between reference frames :math:`N`
+and :math:`A`, it is invertible, and its inverse is equal to the transpose, as
+shown above in the simple example. The direction cosine matrix found in the
```
wwolfie, line 314:
```suggestion
shown above in the simple example. The determinant of the matrix is also always 1, to ensure both associated frames are right-handed. The direction cosine matrix found in the
```

orientation.rst
```diff
@@ -517,52 +621,90 @@ The inverse is also defined on ``A``:
...
-:math:`C`, we can use three successive simple orientations to go from :math:`A` to
-:math:`D`. Using the same technique for the successive simple orientations above,
-but now managing the three dimensional orientations, we can formulate the
-direction cosine matrices for the reference frames.
+If we introduce two additional auxiliary reference frames, :math:`B` and
+:math:`C`, attached to the intermediate camera frame members, we can use three
+successive simple orientations to go from :math:`A` to :math:`D`. We can
+formulate the direction cosine matrices for the reference frames using the same
+technique for the successive simple orientations shown in :ref:`Successive
+Orientations`, but now we will have three dimensional orientation between
```
wwolfie, line 705: A bit unclear, maybe like this: 

```suggestion
Orientations`, but now we will have a sequence of three rotations  between
```

orientation.rst
```diff
@@ -517,52 +621,90 @@ The inverse is also defined on ``A``:
...
+:math:`A` and :math:`D` allowing :math:`D` to be oriented in any direction
```
wwolfie, line 706: Not sure that it is obvious that three is the right number, so would split:
```suggestion
:math:`A` and :math:`D`. With three rotations, :math:`D` can be oriented in any direction
```

orientation.rst
```diff
@@ -662,22 +805,24 @@ handgrip frame :math:`A`:
 
    D.dcm(A)
 
-With these three orientations the camera can be oriented arbitrarily relative
-to the handgrip frame. These successive :math:`z\textrm{-}x\textrm{-}y`
-orientations are a standard way of describing the orientation of two reference
-frames and are often referred to as `Euler Angles`_ [#]_.
+With these three successive orientations the camera can be oriented arbitrarily
```
wwolfie, line 808: Seems slightly redundant. Perhaps rotation is a better word here?

orientation.rst
```diff
@@ -662,22 +805,24 @@ handgrip frame :math:`A`:
...
-the reference frames (or bodies), thus "body fixed orientations". The method
-:external:py:meth:`orient_body_fixed()
-<sympy.physics.vector.frame.ReferenceFrame.orient_body_fixed>` can be used to
-establish the relationship between :math:`A` and :math:`D` without the need to
-create auxiliary reference frames :math:`B` and :math:`C`:
+There are 12 valid sets of successive orientations that can arbitrarily orient
+one reference frame with respect to another. We will also refer to these 12
+possible orientation sets as "body fixed orientations". As we will soon see, a
+rigid body and a reference frame are synonymous from an orientation perspective
+and each successive orientations rotates about a shared unit vector fixed in
```
wwolfie, line 820: typo: orientation singular

```suggestion
and each successive orientation rotates about a shared unit vector fixed in
```

orientation.rst
moorepants: I had a mixture of "rotation" and "orientation" last year when I first wrote this, but ended up changing everything to "orientation" to be consistent and to match the SymPy method names.

Peter230655:
Would this not be called an orthonormal matrix, if A * A.T = I and det(A) = 1 ?

orientation.rst
moorepants: I think it is correct, at least per my definition of the rotation matrix notation.

orientation.rst
moorepants: I tried to reword this:

```rst
successive simple orientations to go from :math:`A` to :math:`D`. We can
formulate the direction cosine matrices for the reference frames using the same
technique for the successive simple orientations shown in :ref:`Successive
Orientations`, but now our sequence of three orientations will enable us to
orient :math:`D` in any way possible relative to :math:`A` in three dimensional
space.
```

moorepants:
Seems like they are the same thing here: https://en.wikipedia.org/wiki/Orthogonal_matrix

orientation.rst
wwolfie: Yes, it is a change in notation that tripped me up

moorepants:
Thanks for the feedback here. Merging.

PR #130, opened 2023-02-16
Added section on alternative representations

wwolfie:
First draft of additional section. Requires some editing.

In terms of content, I left out the actual rules for using quaternions to rotate vectors, as I think it would take too much space for something that is not used in the rest of the materials.

### DIFF
--- a/orientation.rst
+++ b/orientation.rst
@@ -926,26 +926,28 @@ frame or body. There are many alternative approaches to representing
 orientations. Three such representations, which will be used throughout these
 materials, were already introduced:
 
-* **Euler angles** themselves, which provides a minimal representation (only 3
+* **Euler-angles** themselves, which provides a minimal representation (only 3
   numbers), and a relatively straightforward way to compute the change in
   orientation from the angular velocity (see :ref:`Angular Kinematics`).
 * the **direction cosine matrix**, which allow easy rotations or vectors and
   consecutive rotations, both via matrix multiplication,
 * the **axis-angle representation** (used in the
-  :external:py:meth:`orient_axis()
-  <sympy.physics.vector.frame.ReferenceFrame.orient_axis>` method), which is
+  :external:py:meth:`~sympy.physics.vector.frame.ReferenceFrame.orient_axis` method), which is
   often an intuitive way to describe the orientation for manual input, and is
   useful when the axis of rotation is fixed.
 
 Each representation also has downsides. For example, the direction cosine
-matrix consists of 9 elements, less efficient then 3 Euler angles.
+matrix consists of nine elements; more to keep track of than three Euler angles.
 Furthermore, not all combinations of nine elements form a valid direction
 cosine matrix, so we have to be careful to check and enforce validity when
 writing code.
 
-One more frequently used representation is based on so called quaternions,
+Learn more
+==========
+
+One more frequently used approach to representing orientations is based on so called `quaternions`_.
 which are like imaginary numbers, but with three imaginary constant. Apart from
-the usual :math:`i`, :math:`j` and :math:`k` are used.  They act as described
+the usual :math:`i`, also :math:`j` and :math:`k` are used.  They act as described
 by the rule
 
 .. math::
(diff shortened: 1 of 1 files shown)


Peter230655:
I like this summary of _orientation_ very much!

orientation.rst
```diff
@@ -716,8 +716,89 @@ The :math:`z\textrm{-}x\textrm{-}z` Euler angles shown in
 
    D.dcm(A)
 
-.. rubric:: Footnotes
 
+
+Alternatives for Representing Orientation
+==========================================
+
+In the previous section, Euler-angles were used to encode the orientation of a
```
moorepants, line 724:
```suggestion
In the previous section, Euler angles were used to encode the orientation of a
```

orientation.rst
```diff
@@ -716,8 +716,89 @@ The :math:`z\textrm{-}x\textrm{-}z` Euler angles shown in
...
+materials, were already introduced:
+
+* **Euler angles** themselves, which provides a minimal representation (only 3
+  numbers), and a relatively straightforward way to compute the change in
+  orientation from the angular velocity (see :ref:`Angular Kinematics`).
+* the **direction cosine matrix**, which allow easy rotations or vectors and
+  consecutive rotations, both via matrix multiplication,
+* the **axis-angle representation** (used in the
+  :external:py:meth:`orient_axis()
+  <sympy.physics.vector.frame.ReferenceFrame.orient_axis>` method), which is
```
moorepants, line 736: This notation is a bit simpler (I've fixed the others):

```rst
:external:py:meth:`~sympy.physics.vector.frame.ReferenceFrame.orient_axis`
```

orientation.rst
```diff
@@ -716,8 +716,89 @@ The :math:`z\textrm{-}x\textrm{-}z` Euler angles shown in
...
+  often an intuitive way to describe the orientation for manual input, and is
+  useful when the axis of rotation is fixed.
+
+Each representation also has downsides. For example, the direction cosine
+matrix consists of 9 elements, less efficient then 3 Euler angles.
```
moorepants, line 741: maybe "; more to keep track of than three Euler angles".

orientation.rst
```diff
@@ -716,8 +716,89 @@ The :math:`z\textrm{-}x\textrm{-}z` Euler angles shown in
...
+
+.. math::
+
+   i^2 = j^2 = k^2 = ijk = -1,
+
+and the notion that inverting the order of any multiplication between the
+symbols flips the sign of the outcome.
+It is straightforward to convert an orientation described by a rotation of
+angle :math:`\theta` around an axis :math:`\hat{a} = \begin{bmatrix}a_x & a_y &
+a_z\end{bmatrix}^\text{T}` to its quaternion form:
```
moorepants, line 759: I don't use this vector notation in the book. Basically unit vectors can just be written as `\hat{u}` and any other vector is `\bar{v}`. If you want the components of the vector you can define them or use `\bar{v}\cdot\hat{u}` to find the scalar. I also haven't introduced vectors yet, so this is a bit premature to use them.

orientation.rst
```diff
@@ -716,8 +716,89 @@ The :math:`z\textrm{-}x\textrm{-}z` Euler angles shown in
...
+components. For a quaternion representing an orientation, this length is always
+one. Quaternions are part of sympy, and can be used as follows:
+
+.. jupyter-execute::
+
+   ax = sm.symbols('a_x')
+   ay = sm.symbols('a_y')
+   az = sm.symbols('a_z')
+
+   quaternion = sm.Quaternion.from_axis_angle((ax, ay, az), theta)
```
moorepants, line 775: I try to do an intersphinx link to the docs in SymPy whenever a new thing is introduced.

This may also be a bit confusing, as I never have shown them API that calls a method from a class definition. I've tried to make both class instantiation and function use just both seem like function use, so I don't have to talk about OO ever. We do use methods, but I don't really ever say what they are in the OO sense.

orientation.rst
```diff
@@ -716,8 +716,89 @@ The :math:`z\textrm{-}x\textrm{-}z` Euler angles shown in
...
+   print("The norm of the quaternion is {}".format(quaternion.norm()))
+   quaternion
```
moorepants, line 777: I've never used this quaternion module in sympy. It isn't necessarily interoperable with sympy mechanics. We have an `orient_quaternion` method that lets you define rotations using quaternions. If we teach this, I think we should at least teach that, but there's another lesson down the road when we get to simulating where they have to deal with the algebraic equation from each equation.

orientation.rst
```diff
@@ -716,8 +716,89 @@ The :math:`z\textrm{-}x\textrm{-}z` Euler angles shown in
...
+
+It turns out that the multiplication rules for (unit) quaternions provide an
+efficient way to compose multiple rotations, and to numerically integrate the
+orientation when given an angular velocity.
+Due to the interpretation related to the angle and axis representation, it is
+also a somewhat intuitive representation.
+However, the integration algorithm needs to take an additional step to ensure
+the quaternion always has unit length.
+For further study on quaternions, the `Wikipedia page on quaternions
+<https://en.wikipedia.org/wiki/Quaternion>`_ and [Vallery2020]_ are good places
```
moorepants, line 787: Move this link to wikipedia on first mention of quaternions.

moorepants:
I pushed to your branch to wrap the text, fyi.

Peter230655:
At the end of the chapters of many math book I read as a hobby, the is a section _Notes._ Here the author hints at stuff, which, say, deepens what was explained in the chapter, but is not needed for the rest of the book. Is this what you have in mind?

moorepants:
Something like that. My basic design of this text has been to (mostly) provide the minimum set of information needed for them to model and simulate a multibody system. The course already has too much information for the number of credits, so I've tried to streamline the content. I also want them to be able to run a simulation as early in the course as possible (one they construct fully them selves). So, I've dropped various non-central topics. But, we of course would like to at least provide them pointers in the direction of the other related and interesting topics. I could either do that in appendices that aren't tied to the minimal content for the course, or we can have these "learn more" or "notes" sections at the end of each chapter that can lead them to the other stuff.

Peter230655:
Of course, I have no idea how much information for how many credits is correct, but I found your text (last year I read every word) to be "lean" in a very positive way. I do not know whether appendices or notes are better, but I do like the idea of giving directions in which way one could go deeper in points of interest.
For example, I have read about Lie groups, but I never made the connection to rotations. I think, it is very good to give such hints.

orientation.rst
wwolfie: Took your suggestion

orientation.rst
wwolfie: Restructured the example a little bit, so the components are not really needed. As a result, it now also shows the dcm for a given quaternion

orientation.rst
wwolfie: Changed the text so it uses `orient_quaternion`, I don't think there is an easy way to spit out the angle-axis representation from a frame orientation, so I gave an example instead.

wwolfie:
Yes, a "learn more" section is a great plan. Tt avoids repeating "this is not part of the rest of the materials", and it makes it easier for people to skip it, if they are not interested.

In some books, learn-more sections are typeset by placing them in for example a (blue) box. This gives some freedom to place them wherever, and give them an informative title. Maybe that is something to consider?

For now, I put the quaternion explanation and the part about Lie groups in the a learn more section at the end, and kept a short section on the alternatives separate.

moorepants:
This looks great! Merging.

PR #132, opened 2023-02-16
Updates to differentation for 2023

moorepants:
- added learning objectives
- added warning about difference in diff and express

### DIFF
--- a/differentiation.rst
+++ b/differentiation.rst
@@ -127,18 +127,18 @@ scalar is differentiated:
 
 .. jupyter-execute::
 
-   dvdax = v.dot(A.x).diff(alpha)
-   dvdax
+   dvdalphaAx = v.dot(A.x).diff(alpha)
+   dvdalphaAx
 
 .. jupyter-execute::
 
-   dvday = v.dot(A.y).diff(alpha)
-   dvday
+   dvdalphaAy = v.dot(A.y).diff(alpha)
+   dvdalphaAy
 
 .. jupyter-execute::
 
-   dvdaz = v.dot(A.z).diff(alpha)
-   dvdaz
+   dvdalphaAz = v.dot(A.z).diff(alpha)
+   dvdalphaAz
 
 We can then construct the vector :math:`\frac{{}^A\partial \bar{v}}{\partial
 \alpha}` from the new measure numbers know that the :math:`A` unit vectors are
@@ -146,7 +146,7 @@ fixed:
 
 .. jupyter-execute::
 
-   dvdalphaA = dvdax*A.x + dvday*A.y + dvdaz*A.z
+   dvdalphaA = dvdalphaAx*A.x + dvdalphaAy*A.y + dvdalphaAz*A.z
    dvdalphaA
 
 SymPy Mechanics vectors have a special
@@ -173,10 +173,10 @@ above.
 
    .. jupyter-execute::
 
-      dvdex = v.dot(B.x).diff(e)
-      dvdey = v.dot(B.y).diff(e)
-      dvdez = v.dot(B.z).diff(e)
-      dvdex*B.x + dvdey*B.y + dvdez*B.z
+      dvdeBx = v.dot(B.x).diff(e)
+      dvdeBy = v.dot(B.y).diff(e)
+      dvdeBz = v.dot(B.z).diff(e)
+      dvdeBx*B.x + dvdeBy*B.y + dvdeBz*B.z
 
    .. jupyter-execute::
 
--- a/differentiation.rst
+++ b/differentiation.rst
@@ -2,8 +2,6 @@
 Vector Differentiation
 ======================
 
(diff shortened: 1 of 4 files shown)


moorepants:
Same here, had to merge. Out of time. If you notice anything in the published version, let me know.

Peter230655:
Sorry about my stupid question: where do I find the pulished version?

moorepants:
Published versions always end up here: https://moorepants.github.io/learn-multibody-dynamics/

Peter230655:
A bit below equation 31 you write:
....... _Note that this implies that v_hat may not be a vector function of the scalar variable q when observed from another reference frame_.
I think, what is says is this : v_hat = v1(q) in reference frame A does **not** imply that v_hat = v1(q) in another reference fram B.
To me _A implies that B may not._.. sounds a bit strange - but of course English is not my native language.

Peter230655:
Above figure 9 you write:
_There are 12 valid sets of successive orientations that can arbitrarily orient one reference frame with respect to another_.
I assume, these are
- the 6 permutations of (123)
- the six additional possibilities of (121), (131), (212), (232), (313), (323)
Would it make sense to list them?
Is there a theoretical reason to prefer one over the other?

Peter230655:
Above equation 30 you write:
_A general quarternion can thus be written in terms of its components q_0, q_i, q_j, q_k_.
Should one mention, that these components are real numbers?

Peter230655:
Below figure 15 you write:
.._if two vectors are perpendicular: v¯⋅w¯=0 if v¯⊥w¯_
I think, it even is _if and only if_, but maybe I am too formalistic.

moorepants:
Thanks. I addressed all of your comments in https://github.com/moorepants/learn-multibody-dynamics/commit/9194bf5a9f4d42000b66943fea25cdb7ff7a230e

Peter230655:
I think, a 'space" is missing ..only I v_bar... Just a typo, I suppose.
if two vectors are perpendicular: v¯⋅w¯=0 if and _only ifv¯⊥w¯_

Peter230655:
Above equation (43) you write:
_the mutually perpendicular unit vectors fixed in A do not change when observed from A._
Should you add something like: meaning d/dq_r (a_hat) = 0. ?

Peter230655:
Below equation (50) you write:

_d/dq_r (v_hat_i) is also a vector..._.
Maybe you could add something like: _as is obvious from e.g. equation (43), (44), etc_  ?

Peter230655:
Below equation (51) you write:

_Second partials in different reference frames do not necessarily commute_:

This may be surprising for engineers, as they are used to that they do. Your link give the conditions under which they commute, and I believe, for functions used in engineering, they are always fulfilled.
Would it make sense to add something like: _as can be seen by explixitly calculating the left and the right hand sides and confirm they are not equal_ ?

moorepants:
> s can be seen by explixitly calculating the left and the right hand sides and confirm they are not equal ?

There is a "TODO" in the lesson that states we need an example. I just haven't made one. If you make a simple one, I can add it.

Peter230655:
Would this useful?

```

.. code:: ipython3

    import sympy as sm
    import sympy.physics.mechanics as me
    
    q11, q12, q13 = me.dynamicsymbols('q11, q12, q13')
    q21, q22, q23 = me.dynamicsymbols('q21, q22, q23')
    
    a1, a2, a3 = sm.symbols('a1, a2, a3')
    
    N, A1, A2 = sm.symbols('N, A1, A2', cls = me.ReferenceFrame)
    A1.orient_body_fixed(N, (q11, q12, q13), '123')
    A2.orient_body_fixed(A1, (q21, q22, q23), '123')
    
    vector = a1*A2.x + a2*A2.y + a3*A2.z
    
    E1 = (vector.diff(q21, N)).diff(q23, A2)
    E2 = (vector.diff(q23, A2)).diff(q21, N)
    E1 == E2

```

Peter230655:
I think, this is better.

```

#!/usr/bin/env python
# coding: utf-8

# In[ ]:


import sympy as sm
import sympy.physics.mechanics as me

q11, q12, q13 = me.dynamicsymbols('q11, q12, q13')
q21, q22, q23 = me.dynamicsymbols('q21, q22, q23')

a1, a2, a3 = sm.symbols('a1, a2, a3', cls = sm.Function)
a1 = a1(q12, q21)
a2 = a2(q12, q21)
a3 = a3(q12, q21)

N, A1, A2 = sm.symbols('N, A1, A2', cls = me.ReferenceFrame)
A1.orient_body_fixed(N, (q11, q12, q13), '123')
A2.orient_body_fixed(A1, (q21, q22, q23), '123')

vector = a1*A2.x + a2*A2.y + a3*A2.z

E1 = (vector.diff(q12, N)).diff(q21, A2)
E2 = (vector.diff(q21, A2)).diff(q12, N)
E1 == E2

```

Issue #97, opened 2022-07-17
Explain why the holonomic constraints don't include bounds on coordinates or bounds on geometry


moorepants:
1. Students often ask things like "Is 0 deg < q1 <= 180 deg a constraint?" but we don't say anything about these and if they are different than the holonomic equality constraints.

2. Take the four bar linkage. There are constraints on the lengths to obtain a valid linkage. The Grashof condition is part of this: https://en.wikipedia.org/wiki/Four-bar_linkage

See #38 for more discussion.

moorepants:
Russ Tedrake's book has some explanation on this. Maybe I can use that for inspiration on my explanation.

moorepants:
I've added a new warning to address this (a bit).

moorepants:
Basic warning about this is now merged.

PR #138, opened 2023-02-27
Update holonomic constraints chapter for 2023

moorepants:
- Added exercise of Watt's linkage.
- Added how to solve the constraints numerically with SymPy.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,4 +28,4 @@ resizesvg:
 	#inkscape --verb=FitCanvasToDrawing --verb=FileSave --verb=FileQuit *.svg
 
 autobuild:
-	sphinx-autobuild -b html --ignore "$(shell pwd)/_build/jupyter_execute/*" . _build/html/
+	sphinx-autobuild -j "auto" -b html --ignore "$(shell pwd)/_build/jupyter_execute/*" . _build/html/
--- a/README.rst
+++ b/README.rst
@@ -262,6 +262,18 @@ This is also encoded in the Makefile:
 
    make autobuild
 
+If errors occur in jupyter-sphinx cells while editing this will always cause
+the entire book to be rebuilt, i.e. it wipes the sphinx cache. If you set an
+environment variable ``CHAPTER`` to the name of chapter's filename (without the
+``.rst`` extension) only the chapter you are working on will be built. For
+example:
+
+.. code:: bash
+
+   CHAPTER=configuration make autobuild
+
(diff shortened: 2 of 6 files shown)


moorepants:
This is ready for review. Peter, it seems that the preview link above works again so you can see the rendered version.

Peter230655:
Yes, working again.

moorepants:
This is one issue that could be addressed now https://github.com/moorepants/learn-multibody-dynamics/issues/97, but I may not have time to do so before the release at 17:00.

moorepants:
I'm merging this. Feel free to give feedback here later.

Peter230655:
There is a discussion going on between Timo and 'wwolfie', whether in your equation (70) point O must be fixed in frame A or not. wwolfie's proof looks reasonable to me, but I am not sure as he introduced a frame F, maybe a typo.
Should it be mentioned there, that point O must be fixed (or not fixed, whatever is right)?

Peter230655:
Above the NOTE: Sympy mechanics has no way of knowing....
you write O.set_vel(N, 0)
Timo once told me, that one should not write O.set_vel(N, 0**.**)
Is this worth mentioning?

Peter230655:
_Note
Sympy mechanics has no way of knowing..._
I think, this is so, because Newton's equations are the same, whether a frame is at rest or in uniform motion.
Is this not even an axiom of Newton's mechanics?
If so, is it worth mentioning?

Peter230655:
Below eq (92) you write:
_We include t as it is possible that the constraint is an explicit function of time (instead of only implicit, as seen above in the four-bar linkage example)._
I guess an example of t being explicit in the constraints would be if we let, say, l_n = l_n * ( 1 + 1/2 * sin( alpha * t))

Peter230655:
Above eq (97) you write:
_Our holonomic constraints dictate that there is no relative motion between points or reference frames_,
There is no relative motion between, say, P1 and P2, but is there not a relative motion possible between P1 and P3, or P2 and P4?

Peter230655:
This is only for my personal education: you write:
_t = me.dynamicsymbols._t_
In my examples, I always use t = sm.symbols('t'), and this seems to work.
What is the advantage of your way of defining t ?
Thanks!

Peter230655:
Above eq (84) you write:
_The case of Fig. 22 there are two vector paths to point P4 from point P1._..
Would it be (very slightly!) better to refer to Fig 22 a.)  ?

PR #140, opened 2023-03-02
[WIP]: Feature/lagrangian

wwolfie:
This PR adds a chapter on the Lagrange method to derive equations of motions.
Currently, it only contains an outline, a first draft of some of the initial sections, and the first example.

The chapter now starts with a kinetic-energy only system, as that seems the fastest way to get to useful
skills given that the readers know about Kane's method and kinetic energy, but not yet about potential energy.

Adding potential energy and adding constraints are the next sections, with two learn more sections at the end.

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -52,6 +52,7 @@ taught by `Jason K.  Moore`_ but it is designed to standalone from the course.
    holonomic-eom
    noncontributing
    tmt
+   lagrange
    notation
    references
 
--- a/lagrange.rst
+++ b/lagrange.rst
@@ -50,7 +50,7 @@ their time derivatives.
 
     In Kane's method, we are free to choose generalized speeds :math:`\bar{u}` that differ from
     the time derivatives of the generalized coordinates, that is :math:`\dot{\bar{q}}`. This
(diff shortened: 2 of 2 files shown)


Peter230655:
Just to make sure, I understood correctly: This will be added as a chapter to Jason"s lecture, correct?
Thanks!

wwolfie:
Just updated the draft. There are still some figures and two small examples missing, but it is probably now useful to look at it.

Peter230655:
I would like to read it, as I know next to nothing about the Lagrangian method.
With Jason's changes I must always wait until he has merged them into the lecture, as before they do not look 'readable' to me. 
Anyway I can get this in a 'readable' way before it is merged?
Thanks & sorry to bother you with such trivial (except to me they unfortunately are not) things!

moorepants:
@wwolfie I could pull this branch and then push a new PR, which would trigger the preview. Then I can make you collab so you can push to it.

wwolfie:
Sounds great!

moorepants:
Moved to #156

@wwolfie  you have collab rights

FYI, I often commit directly to master, so beware.

Issue #139, opened 2023-02-27
Implicit assumption on the velocity calculation in one of the equations


tjstienstra:
I got an e-mail of two students, who failed one of the multiple choice questions by miss interpreting equation 70 of the translation kinematics chapter. [This equation](https://github.com/moorepants/learn-multibody-dynamics/blame/master/translational.rst#L53) assumes that point O is also fixed in reference frame B, while that is not mentioned explicitly in the text.

Peter230655:
Stupid question: is O not really a function, a constant function so to speak?
Then in would not change in any frame as I understood it?

TJStienstra:
O is a point.

Peter230655:
I agree, of course. But is a point equal to a vector?
I am unsure whether when I think about a point, do I subconcientiously think about a vector?

Peter230655:
If O was (implicitly) assumed to be the origin of A, then it would be fixed in any other frame?

TJStienstra:
A point is a location and not equal to a vector. However to say something about a point, you kind of always need a reference point with respect to which you are saying something, so you'll actually get a vector. If you by example have the following equation:
$${}^A\bar{v}^P = \frac{{}^A\bar{r}^{P/O}}{dt}$$
This equation says that the velocity of point P observed from frame A is defined as the derivative with respect to time of the vector from O to P, where O is an arbritrary point fixed in A. You can say O is an origin of A, but actually every point is, as long as it has zero velocity in A. This point O however can have a different velocity in the B frame (or be undefined as you know from sympy).

Peter230655:
_A point is a location and not equal to a vector. However to say something about a point, you kind of always need a reference point with respect to which you are saying something, so you'll actually get a vector._ 
 This makes emminent sense, thanks for putting it this way!

So, eq (70) in Jason's script would be correct for any O on the line given by omega, the angular velocity vektor of the rotation of B w.r.t. frame A? If true, it should be correct in particular if O is the origin of A?

Peter230655:
A while back, I raised issue #24669, where I got the suspicion, that there may be an issue with v2pt_theory:
A ball rolls on a street without slipping or sliding or bouncing, the street being a line in the X / Y plane
If I apply v2pt_theory for the center (Dmc) of the ball, and with the contact point (assumed) momentaryly fixed in the body frame of the ball, I get incorrect results ( total energyis not constant), unless the street is a horizontal line.
[...]

PR #141, opened 2023-03-02
Fixing notation and implicit assumptions in translational velocity equation 

wwolfie:
[Fix for issue 139](https://github.com/moorepants/learn-multibody-dynamics/issues/139)

Explanation in [this comment](https://github.com/moorepants/learn-multibody-dynamics/issues/139#issuecomment-1452236280)

### DIFF
--- a/translational.rst
+++ b/translational.rst
@@ -30,14 +30,14 @@ and accelerations of points in a system.
 Translational Velocity
 ======================
 
-If a point :math:`P` is moving with respect to a point :math:`O` that is fixed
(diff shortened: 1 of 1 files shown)


TJStienstra:
I would not suggest doing these changes as it conflicts the notation used in general, which is correct. The zero velocity point is implicit and can be any point which is not moving in the frame. The change should either be mentioning that O is also assumed to be fixed in B or to update the equation to the one point velocity theorem. The latter however would complicate things more than reasonable.

PR #144, opened 2023-03-08
Improve and correct coordinate notation and counting

### LINKED ISSUE #143: Improve the precision of the wording about what is an isn't a generalized coordinate
In the holonomic chapter I need to state that there can be N coordinates and only n=N-M generalized coordinates. I am a bit too loose with the wording around this. Also add some parts to the notation page to show what N, M, n, m, p are.

moorepants:
Fixes #143

- Corrected the holonomic constrain equation to use N instead of n, where N is the number of coordinates.
- Added n = N - M equation.
- Added a notation explanation on N, M, n, m, p.

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -183,26 +183,28 @@ the intuitive choice being :math:`\hat{n}_x` and :math:`\hat{y}`.
 
 For the loop to close, these two expressions must equal zero for all values
 :math:`q_1,q_2,q_3`. These are two nonlinear equations in three time varying
-variables. The solution can be found if we solve for two of the time varying
-variables. For example, :math:`q_2` and :math:`q_3` can be solved for in terms
-of :math:`q_1`. We would then say that :math:`q_2` and :math:`q_3` depend on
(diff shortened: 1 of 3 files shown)


configuration.rst
```diff
@@ -183,40 +183,44 @@ the intuitive choice being :math:`\hat{n}_x` and :math:`\hat{y}`.
...
+:math:`q_3` depend on :math:`q_1`. These two equations are called holonomic
+constraints, or configuration constraints, because they constrain the kinematic
+configuration to be a loop. Holonomic constraints take the form of a real
+valued vector function:
 
 .. math::
    :label: configuration-constraint
 
-   \bar{f}_h(q_1, \ldots, q_n, t) = 0 \textrm{ where } \bar{f}_h \in \mathbb{R}^M
+   \bar{f}_h(q_1, \ldots, q_N, t) = 0 \textrm{ where } \bar{f}_h \in \mathbb{R}^M
```
moorepants, line 197: I had this correct in the notation page, but it is incorrect here.

PR #145, opened 2023-03-08
Updates to the motion constraints (noholonomic) chanter for 2023.

moorepants:


motion.rst
```diff
@@ -218,16 +218,44 @@ We see that to for the last two pairs, the mixed partials do not commute. This
 proves that :math:`f_n` is not integrable and is thus an essential nonholonomic
```
WouterWolfslag, line 218: is essential correct here?

moorepants: Maybe not the best word. I want mean that the nonholonomic constraint is an irreducible nonholonomic constraint and that you have to consider it as a nonholonomic constraint.

moorepants: It is essential in the sense that you can't transform it into a holonomic constraint and reduce the number of generalized coordinates. It will reduce the number of degrees of freedom but not by reducing the number of generalized coordinates.

Peter230655:
1.
Just curiosity: Is the the criteria that the mixed partials commute _necessary_ or _necessary and sufficient_ for the integrability?
2.
In the chapter holonomic constraints, you said that only in very are cases can the holonomic constraints be solved symbolically for the dependent coordinates. Does this not mean, that the holonomic constraints most of the time cannot be used to reduce the number of general coordinates?
3.
Would one not normally know from 'physical' consideration, whether a constraint is holonomic or non-holonomic?

Peter230655:
You write in a note:
_We could find a very strong person to push the car sideways, overcoming the very high resisting friction force. It is important to note that **any constraint is just a model of a physical phenomena**. We know that if we push hard enough and low enough that the car’s lateral motion is not constrained. Also, if the car were on ice, then the nonholomonic constraint would be a poor modeling decision_.

Is this not true for everything we are doing in mechanics?
- mass points do not exist in reality
- rigid bodies do not exist in reality
- Hooke's law is an approximation
- etc..
But these are idealisations, which work well, and allow us to do calculations?

Peter230655:
Above eq(102) you write:
_We are working towards writing the equations of motion of a multibody system, which will be differential equations in a first order form_
I guess, one reason is, that numerical solvers for ODEs always solve systems of first order differential equations only.
Are there other reasons for doing this?

moorepants:
> Is this not true for everything we are doing in mechanics?

Yes, but I think we are much less likely to question some modelling assumptions than others. Students don't seem to question adding configuration constraint equations to model a real bearing as an ideal revolute joint or modeling a steel bar as a cylindrical rigid body with uniform density. Whether we should model a car tire to ground interaction in parallel parking as a nonholonomic constraint should cause some questioning of the assumption. I guess my wording in this chapter is to try to promote that these constraint models we invent aren't infallible, and it is easier to see that they aren't because nonholonomic constraints are more easily violated in real life than the kinematic constraints we've already spoke of. But maybe I assume to much here. I don't really know how well the students know why we make certain modeling decisions. I'm not explicitly teaching how to model in this course, only hoping that it is gained implicitly in some manner.

Peter230655:
I have never taught in my life, I guess your experience is the best guide of what to mention and what  not. Intuitively, I think I can comprehend why non-holonomic constraints are more likely to be questioned.

moorepants:
A problem with teaching is that I assume that students will make the same thought paths, conclusions, and assumptions as I did when I first learned these things. All that is reflected in what I write and lecture. But there isn't really isn't an easy way to know if how I see it is how others see it.

moorepants:
One example, is that I didn't deeply understand multibody dynamics until I had coded the fundamental algorithms myself. Reading the math in Kane's book only got me so far. So this whole text "learn multibody dynamics" is me writing how I would have wanted to learn it if I went back to 2005 and did it over again. In some sense, I can only hope that this is a good approach for most of my students too. I've had at least a couple students tell me they'd rather have more mathematical presentation.

Peter230655:
I am a bit mathematically inclined - and I do not see where your lecture would lack in rigor at all! Quite the opposite in my opinion.
And the examples you do in your lecture, we simply could not do 45 years ago when I learned it: there was no sympy mechanics around.

moorepants:
I had to merge this, out of time. I'll consider the other comments here and proofread tonight and tomorrow.

motion.rst
```diff
@@ -164,10 +178,10 @@ time we can retrieve the original holonomic constraint, so it really isn't a
 nonholonomic constraint even though it looks like one.
 
 So if we can integrate :math:`f_n` with respect to time and we arrive at a
-function of only the generalized coordinates and time, then we do not have an
-essential nonholonomic constraint, but a holonomic constraint in disguise.
-Unfortunately, it is not generally possible to integrate :math:`f_n` so we can
-check the integrability of :math:`f_n` indirectly.
+function of only the generalized coordinates and time, then we do not have a
+nonholonomic constraint, but a holonomic constraint in disguise. Unfortunately,
+it is not generally possible to integrate :math:`f_n` so we can check the
```
wwolfie, line 183: I think what this is meant to convey is that even if it is possible to integrate f_n, it might be hard to do so. So trying to integrate and failing,  is a poor check for integrability. I think that is now a bit tricky to understand from the sentence. 

Suggestion:

Unfortunately, directly integrating :math:`f_n` is often difficult, even if it is possible. Therefore we use a method to check the integrability of :math:`f_n` indirectly.

motion.rst
```diff
@@ -216,24 +230,152 @@ and the other two pairs:
...
+
+.. math::
+
+   \dot{q}_1 - \dot{q}_2 r = 0
+
+This is a scalar nonholonomic constraint equation that ensures rolling without
+slip. Take care to calculate the relative velocities of the two points fixed in
+each of the bodies in rolling contact that are located at the contact point at
+that *instance of time*.
+
```
wwolfie, line 360: Maybe add a note here that while rolling without slip is often essentially/truly non-holonomic, in this 2D case, the constraint is integrable?

moorepants: I've never thought about this. This violates the whole explanation I have about integrability. Is it because the 2D rolling without slip is essentially creating an instantaneous pin join at the contact point?

moorepants: I guess this is just a holonomic constraint for the 2D case.

moorepants: I fixed it to say it is holonomic.

moorepants: I think that I have thought only about the 3d rolling case so much, that I forgot this reduces to holonomic. Thanks for finding the mistake!

Peter230655:
Is the 2D case integrable, when the plane N in fig 29 is replace by a smooth line, say y = f(x) ?
- $\frac{d}{dt}(x(t)) = \dfrac{-(r \cdot u)} {\sqrt{1 + \left(\frac{d}{dx}(gesamt(x(t)\right)^2}}\$

moorepants:
I haven't worked out a non-straight surface, but my intuition is that you have to determine the geometry of the two points in contact and calculate their velocities, same as I explained here. The geometry problem of determining the position vector to the point fixed in the disc that is located at the contact point is typically a tricky part.

Peter230655:
I think

$\frac{d}{dt}(q_1(t)) = \dfrac{-(r \cdot \dot q_2)} {\sqrt{1 + \left(\frac{d}{dx}(gesamt(x(t)\right)^2}}\$

Peter230655:
On my rolling ellipse on an uneven street, the term $f_n$ is:

$\dot q_1(t) - \dfrac{-\sqrt{a^2sin(q_2(t))^2 + b^2cos(q_2(t))^2}} {\sqrt{1 + \left((\frac{d}{dq_1}(gesamt(q_1(t))\right) ^2}} \cdot \dot q_2(t)$

Peter230655:
Above eq (109) you write:
_These laws of motion are second order differential equations._.
This is 'dictated' by nature, not mathematics.
Would it make sense, to add a sentence to this respect? When I studied years ago, it took me a while to distinguish what is mathematics, what is dictated by physics.

Peter230655:
Below fig(31) you write: The velocity of $A_0$ in N is a simple time derivative, and below there is the first line in the code:
O.set_vel(N, 0)
Would it make sense to refer to that modern art example, where you explained, that with any such problems, the speed of one point must be set?
I never know, how much repetition is too much repetition.

Peter230655:
Below equation (119) you write:
_Because nonholonomic constraints are derived from measure numbers of velocity vectors, the nonholonomic constraints are always **linear** in the generalized speeds_.
Is this linearity not a consequence of this rule of differentiation: $\frac{d}{dt} f(x(t)) = \frac{d}{dx} f(x) \cdot \dot x $

Peter230655:
Above eq(122) you write:
_For simple nonholonomic systems observed in a reference frame A_
In fig(31), and, I believe, quite often, you call the reference frame **N**.
Would it be better to call it N? (A VERY minor point, of course)

Peter230655:
Below eq(122) you write:
_If there are no nonholonomic constraints, the system is a holonomic system in A_.
This is the definition for a holonomic system, I assume.
For my understanding only: for a system to be holonomic, there is no need for explicit holonomic constraints to be present?
Or is the fact that the snake board cannot take off the ground an implicit holonomic constraint?

PR #146, opened 2023-03-15
Update first half of the mass chapter for 2023

moorepants:


Peter230655:
Above eq(123) you write:
_Each point in the distribution of mass can translate_.
Could this sentence be misunderstood to mean that points of mass can translate relative to other points of mass of the _same_ rigid body? I think, in a rigid body they cannot.

Peter230655:
Above eq(124) you write:
_For a solid body with a density ρ....._
A few lines above, you called the same thing a _rigid_ body.
Would it make sense to use the same word 'rigid' here, too?

Peter230655:
Above eq(125) you write:
the first moment of mass _can_ be defined as..
Does this imply, that there are other definitions?

Peter230655:
The proof that eq(127) gives the right point $S_o$ could be proven by $\bar r^{S \ P_i} =

Peter230655:
Is eq(130) totally correct?
|$\bar I_a| ge 0$

mass.rst
```diff
@@ -258,13 +315,92 @@ rigid body. The radius of gyration about a line through :math:`O` parallel to
 
    k_{aa} := \sqrt{\frac{I_{aa}}{m}}
 
+.. todo:: Double check the solution to this exercise!
+
+.. admonition:: Exercise
```
moorepants, line 320: I'm not sure I have the correct answer to this new exercise. If someone else has time to have a look, that would be helpful. I'll revisit after the other edits.

moorepants:
> Is eq(130) totally correct?

I think that angle will be between 0 and pi/2. Need to clarify if so.

moorepants:
> I think that angle will be between 0 and pi/2. Need to clarify if so.

I made it |sin(theta)|. I think that is correct.

mass.rst
moorepants: I think it is correct now.

Peter230655:
Maybe in eq(124) you should change the domain of the integration from _solid_ to _rigid body_ or similar?

moorepants:
Actually I think I wanted to use solid there, not rigid. I'm not intending to talk about the concept of a rigid body, but simply a solid with some volumetric boundary.

Peter230655:
I think I understand. These equations hold true even if the body is not rigid.

Peter230655:
In equation (136), there is a term $vector^2$
Should you mention, that $vector^2$ := dto($vector, vector$)?
( I guess, it is pretty obvious from the context )

Peter230655:
In the solution to the excercise below eq(137) you write:
_Here we see that rotating the set of points around the ring will maximimize and minimize the radius_
- I guess, you mean rotation point O around the ring?
- _maximimize_, this is misspelled, I think.
NB: A very neat excercise, I think!

Peter230655:
Above eq(138) you write:
_..(6 unique scalars)..._
Would it make sense to write: ...(6 unique scalars, _because of eq(135)_)... ?
As I said, I never know, how much repetition is too much repetition.

Peter230655:
Somewhat below eq(141) you write:
_Unit dyads are analogous to unit vectors. Here is another example_:
I wonder, whether you should say here a few more words. This dyad space is nine dimensional, hence nine unit 'vectors' are needed, $\hat a_i \bigotimes \hat b_j$, where the $\hat a_i, \hat b_j$ are unit vectors in frames A, and B, for i, j in (x, y, z).

Peter230655:
Under Properties of Dyadics you write:

Left and right dot product with a vector (results in a vector):

u¯⋅(v¯⊗w¯)=(u¯⋅v¯)w¯
u
¯
⋅
(
v
¯
⊗
w
¯
)
=
(
u
¯
⋅
v
¯
)
w
¯

(u¯⊗v¯)⋅w¯=u¯(v¯⋅w¯)
(
u
¯
⊗
v
¯
)
⋅
w
¯
=
u
¯
(
v
¯
⋅
w
¯
)

Left and right cross product with a vector (results in a dyadic):

u¯×(v¯⊗w¯)=(u¯×v¯)⊗w¯
u
¯
×
(
v
¯
⊗
w
¯
)
=
(
u
¯
×
v
¯
)
⊗
w
¯

(u¯⊗v¯)×w¯=u¯⊗(v¯×w¯)
(
u
¯
⊗
v
¯
)
×
w
¯
=
u
¯
⊗
(
v
¯
×
w
¯
)

Peter230655:
dyadic P˘
P
˘

Peter230655:
Above eq(148) you write:
_This inertia dyadic can easily be expressed relative to another reference frame if the orientation is defined_:
I believe, here you use the relationship between frames A and B as defined above eq(142)

Peter230655:
The left hand sides of eq(149) andwq(150) are vectors, as is clear from your definition of the dot product between a vector and a dyad further up in the text. Also calling in $\bar H$ makes it clear.
Does mit make sense to still say it, like The angular momentum.......is definded as the vector: ?

Peter230655:
Above eq(153) you write:
_The consequence of $\bar I_a$ is parallel to $\hat n_a$ is that the products of inertia are all zero_
I believe, this is an immediate consequence of eq(133): if $\bar I_a \parallel \hat n_a$, then $ \bar I_{ab} \cdot \ n_b$ = 0

moorepants:
I addressed about half or more of your comments in this commit: https://github.com/moorepants/learn-multibody-dynamics/pull/147/commits/1cd38f6f8a4211c163815f4098c7ea1e8236bbff

I didn't add some of the notes, as I think it is worth that the students try to think these things out themselves.

PR #147, opened 2023-03-23
Update second half of mass distribution chapter for 2023

moorepants:


Peter230655:
Will you inform, when it becomes part of the lecture, so it is in 'nice' print?
Thanks!

Peter230655:
Above eq(109) you write:
_differential equations that are most useful for simulation when in a first order form_.
I assume 'simulation' means numerical integration.

Peter230655:
Parallel Axis Theorem
_If you know the central inertia dyadic of a rigid body B (or equivalently a set of particles) about its **mass center** Bo
then it is possible to calculate the inertia dyadic about any other point O_
Does $B_o$ have to be the mass center, or would the theorem also hold for any other point?
(Of course, it makes paractical sense to take $B_o$ as the mass center)

Peter230655:
In the excercise below eq (150) you write:
_Gvine the inertia..._
I believe _Gvine_ is a typo.

Peter230655:
In the excercise below er (150) you write:
..._the head tube angle is 68 degrees from the ground plane_.
Is it obvious what this angle is, or would a sketch help?

Peter230655:
https://aapt.scitation.org/doi/10.1119/1.4994835
I think, this link gives a generalization of the Huygens Steiner theorem - but I did not study it

PR #150, opened 2023-03-29
Update loads.rst for 2023

moorepants:


Peter230655:
Force, Moment and Torque.
Right below the heading you write:
_A force is an abstraction we use to describe something that causes mass to move (i.e. accelerate from a stationary state)_.
A stationary state is a state of rest or of uniform motion, I believe. Would students easily see, that uniform motion is also a stationary state?

Peter230655:
Above fig (33) you write:
_Angular velocity is an example of a free vector. It has a direction and magnitude, but is not associated with any line of action. A force vector, on the other hand, is bound_.
While it is intuitively clear, at least to me, that a force is a bound vector, it is much less obvious to me that angular velocity is a free vector. This 'picture' has helped my intuition:
I have a plate which I can rotate around a fixed point. No matter where I put a body on this plate, if I rotate the plate with $\dot \phi$ the body will rotate with the same $\dot \phi$. This shows that the vector $\dot \phi$ must be free. No idea if useful, at least it helped me.

Peter230655:
In eq(156), the $\cross$ operation ensures, that the moment $\bar M$ does not depend which point on the line ($L_1, L_2...$ is taken.
I never know, whether it is worth mentioning somethnig like this.

moorepants:
> I have a plate which I can rotate around a fixed point. No matter where I put a body on this plate, if I rotate the plate with
the body will rotate with the same

I added a variation on this. Thanks.

> In eq(156), the operation ensures, that the moment

Also added.

PR #151, opened 2023-03-31
Updated the collision section (connected to Hunt-Crossley model)

moorepants:


Peter230655:
Below eq(157) you write:
_....moments due to each vector about P is call the..._
I think, there is a typo, should it not be ....is call**ed**...?

Peter230655:
In fig (37) would it make sense to inficate $\hat w_x$?
Of course, it is clear from the definition below, and maybe it will 'overload' fig (37)?

moorepants:
> In fig (37) would it make sense to inficate $\hat w_x$?

I'm not sure what that is? I don't see anything missing.

moorepants:
Nevermind, I see. The unit vector for the wheel frame. Yes, that would help.

Peter230655:
I think eq(161) is called a _line integral_ in mathematics, at least in German it is called Linienintegral.

Peter230655:
Below eq(161(, you give this example of non contributing forces:
_contact forces on particles across smooth (frictionless) surfaces of rigid bodies_
Frankly, it took me a while to get it. I guess the explanation is this:
- in the tangential plane of the contact there are no forces, as the friction is zero
- normal to the tangential plane, there is a contact force, but it is non contributing since the body is rigid, that means the path S is zero.

Peter230655:
In fig (39) you write _seperated_ 
I believe it is sep**a**rated, but my spelling skills are lousy, both in English and German.

Peter230655:
In eq (164), you give $C_d$.
Dies this correspond to the $c_v$ value, which car magazines report on cars, the smaller the value the more "airodynamic' the car is?

Peter230655:
A bit below eq(168) you write:
_The impact force model is often combined with a friction model to generate a friction force for impacts that are not perfectly normal to the contacting surfaces. For example, Coulomb friction force can slow the particle’s sliding on the surface if we know the tangential velocity components vx and vy at the contact location_.
I believe, this Coulomb friction force is equivalent to a force on the center of gravity of the colliding body, and a torque on the body. Since you explained this equivalence earlier, maybe worth pointing out here? (If indeed I am right about this. At least I used in in my model of colliding billiard balls)

moorepants:
> I believe, this Coulomb friction force is equivalent to a force on the center of gravity of the colliding body, and a torque on the body. Since you explained this equivalence earlier, maybe worth pointing out here? (If indeed I am right about this. At least I used in in my model of colliding billiard balls)

I intended to describe the forces in the final sections as forces that act on any point. The forces can all be resolved as a resultant at the mass center and torque of a couple acting on the body.

moorepants:
> Dies this correspond to the value, which car magazines report on cars, the smaller the value the more "airodynamic' the car is?

Possibly, but I don't know of a $c_v$. They may be reporting $C_d A$ which is the combined value giving a number for a specific car.

moorepants:
>  I believe it is separated, but my spelling skills are lousy, both in English and German.

Fixed, thanks. Apparently this is one of the most misspelled words in English :)

Peter230655:
Below eq(166) you write:
_n is typically 3/2_,
If I understood [Flores2023] correctly, $n = \frac{3}{2}$ was derived theoretically by Hertz?

moorepants:
Yes, I should say that instead of my wording.

wwolfie:
Some belated comments on this PR:

**Line 292**
_"Forces are bound vectors that can be considered acting on specific points, thus
we will always need a vector and a point to fully describe the force"_

The attachment point is just the most convenient way to help describe the line (especially when the direction of the Force changes over time). Maybe it is more precise to say: _"Forces are bound vectors, so we have to define their lines of action. This is best done by specifying the points on which each force acts"_

**Line 377**
_"Contributing forces are those that do work on the multibody system."_

Maybe it is better to add "can/potentially" to this sentence. If the system happens to move perpendicular to a force for a while, due to a coincidence involving other forces, but is not constrained to to so, I think the force should be considered contributing. If we know the path to be constrained (e.g. level ground wrt gravity),  the force is non-contributing.

Similarly, non-contributing forces  _can never_ do work.

**Line 403**
Contact without slip is not a special case for frictionless contact. I'd say infinite (or at least sufficiently high) friction is the model there.

**Line 647**
Maybe add a line: _Such contact force models can provide more accurate results, at the cost of much longer computation times_

**Line 660**
Hysterias ==> Hysteresis

**Line 713**
The part with \dot{z} if z <0, 0 otherwise is strictly speaking not necessary, as the term in front of it will be zero if z > 0.
This holds for the next functions too. Maybe be more explicit in Line 651 if you decide to remove it: "This difference between the absolute value and the value itself is equivalent to"

moorepants:
Thanks guys, I included all your feedback here: f872112

PR #153, opened 2023-05-01
Update equation of motion chapter for 2023

moorepants:


moorepants:
I'm late on this one, so just going to merge. If you notice any errors I'll fix after merging.

Peter230655:
Above eq(186) you write:
_This example may seem overly complicated when using Kane’s method, but it is a systematic method that works for any number of rigid bodies and particles in a system_
I assume, the purpose of Dr. Kane was to create a system, which allows to establish the equations of motion in a 'formalistic' way (and allowing to ignore all internal forces!), eventually leading to something like sympy mechanics.
If I am right, would it make sense to mention something like this?

Peter230655:
I assume the chapter Equations of Motion, eq(188) - equ(190) is to convert a system or _second order_ differential equations into a system of _first order_ differential equations, because available routines like solve_ivp only solve system of first order differential equations.
If I am right, worth mentioning this purpose?

Peter230655:
Below eq(187) you write:
_Note that we use the var_in_dcm=False keyword argument. We do this because the generalized speeds are not present in the unspecified direction cosine matrix relating A and B._
Just so I understand correctly: You could have specified an orientation between A and B and skipped var_in_dcm=False, result would be the same?

Peter230655:
Warning below fig 47:
I take it, the save thing is to put all substitutions into one dictionary, rather than doing like A.subs(dict_1).subs(dict_2)?

PR #155, opened 2023-05-11
Update the nonholonomic equation of motion chapter

moorepants:
Minor update for this chapter.

### DIFF
--- a/nonholonomic-eom.rst
+++ b/nonholonomic-eom.rst
@@ -588,7 +588,7 @@ later use in the animation of the motion.
 
 .. jupyter-execute::
 
-   t0, tf = 0.0, 1.0
+   t0, tf = 0.0, 8.0
 
 .. jupyter-execute::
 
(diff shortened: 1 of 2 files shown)


Peter230655:
Below the heading **Numerical Evaluation** is a Warning.

There you write:
_sym_mat = sm.Matrix([[a, b], [c, d]])
eval_sym_mat = sm.lambdify((a, b, c, d), sym_mat)
num_mat = eval_sym_mat(1.0, 2.0, 3.0, 4.0)
num_mat_

Would it make sense to add cse=True?
I guess with this example it does not matter, but I think, 50% of my examples would take 'forever' to integrate without this.

moorepants:
That is a different chapter than this, but I think I should use cse=True in all lambdify calls after the SymPy chapter (where I originally introduce lambdify).

Peter230655:
Sorry, I still get confused, where to comment!

PR #156, opened 2023-05-11
Adds chapter on the Lagrange approach

moorepants:


moorepants:
@wwolfie maybe this needs a master merge or rebase?

moorepants:
Getting a sphinx error:

```
Exception occurred:
  File "/home/runner/micromamba-root/envs/multibody-book/lib/python3.8/site-packages/jupyter_sphinx/ast.py", line 46, in load_content
    raise OSError(f"File {filename} not found or reading it failed")
OSError: File /home/runner/work/learn-multibody-dynamics/learn-multibody-dynamics/# Setting up reference frames
psi,theta, phi, x, y, z = me.dynamicsymbols('psi theta phi x y z')
N = me.ReferenceFrame('N')
B = me.ReferenceFrame('B')
B.orient_body_fixed(N, (psi, theta, phi), 'zxy') not found or reading it failed
```

lagrange.rst
```diff
@@ -0,0 +1,518 @@
...
+
+   p1, p2, p3 = me.dynamicsymbols('p1, p2, p3')
+   p_sym = sm.Matrix([p1, p2, p3])
+   J_p_wrt_qdot = p.transpose().jacobian(q_dot)
+   p_dot = rhs - L.jacobian(q).transpose()
+   q_dot_solve = J_p_wrt_qdot.solve(p_sym)
+
+There are two important realizations:
+
+.. jupyer-execute::
```
moorepants, line 448:
```suggestion
.. jupyter-execute::
```

wwolfie:
Ouch, a bit of a mess. I think there was an issue with the indentation settings in my IDE.

Made some more fixes, and expanded the constrained-system section.

moorepants:
I was getting major merge conflicts and couldn't figure it out. Now I see you force pushed. I'm not sure how to pull things down now other than deleting my local branch. We probably shouldn't force push when more than one person is working on a branch.

moorepants:
Do you know what I should do locally to get in a working state? I'm a bit stuck.

moorepants:
I think I got it!

moorepants:
> Ouch, a bit of a mess. I think there was an issue with the indentation settings in my IDE.

note that I've been using 3 different indentation patterns: https://github.com/moorepants/learn-multibody-dynamics/#editing-guide which may not be standard in IDE config files (I don't have it working in vim).

Peter230655:
Under the heading Introduction you write:
_This book has already discussed **three** methods to derive the equations of motion of multibody systems: Newton-Euler, Kane’s method and the TMT method. This chapter will add a **third:** the Lagrange method, originally developed by Joseph-Louis Lagrange_
I believe, it should be: a _forth_ method.

Peter230655:
The language in this chapter (Lagrange) is a bit different from earlier chapters.
For example, here **we** is used frequently, as in _**we** have learned earlier._..
Unimportant in my opinion, merely an observation.

Peter230655:
Under Example: _3D body in outer space_, you give the code on how to calculate it.
In previous chapters, the results were also given, I think, it would be nice to give them here, too.

I guess, the results would also explain the sentence right below the code:

...but the terms, particularly the terms involving $\dot{\dot q_2} are mangled.

Peter230655:
It seems that not all the links are working.
_...is done for a system of point-masses in [Vallery2020]._
This link for example does not seem to work with me.

_Some applied forces, known as conservative forces conservative forces, can be._...
This link worked with me.

Peter230655:
Below eq(228) you write:
_Example: 2D double pendulum with springs and sliding pointmass_
Would this not be a **3**D example, the example of fig 47?

Peter230655:
Under the heading **Constraint equations of motion** you write:
_This is not possible for non-holonomic constraint (by definition), and... _
I guess, if the non-holonomic constraints are really 'holonomic constraints in disguise' (in a previous chapter it was shown how to check this) then, by definition, this is not a non-holonomic system?

Peter230655:
Under the heading **Lagrange's vs Kane's** you write:
_The Lagrangian method is the **second** method to deriving the equations of motion presented in this book, after Kane’s method_.
This seems to contradict the sentence at the very top of the lecture, where you call Lagrange's method the forth method?

index.rst
```diff
@@ -52,6 +52,7 @@ taught by `Jason K.  Moore`_ but it is designed to standalone from the course.
    holonomic-eom
    noncontributing
    tmt
+   lagrange
```
moorepants, line 55: Let's put this before TMT, to match the flow of the course. I'll put energy/power before lagrange.

moorepants:
>  For example, here we is used frequently, as in we have learned earlier...

We have a new co-author here, so there will be some style differences.

Peter230655:
In eq(233), maybe more consistent to write $a_r(\bar q)$, as in eq(231) and in eq(232) ?

Peter230655:
Beloq eq(233) you write:
_..the velocity of the contact point (G) is zero in **both** the n^x, n^y and n^z direction._
Would _both_ not refer to two items, rather than to three like here? 
Of course, German is my native language, so I may easily be worng.

Peter230655:
eq(234) looks slightly 'odd' to me:
Would it look slightly nicer to put -rn_x in brackets like (-rn_x)?
Of course very minor!

Peter230655:
Above eq(238) you write:
_This statement is a particular case of the so called Noether’s theorem_.

In previous lectures you had links connected with the statement, pointing to some article in the internet about the statement.
Surely Ms. Emmy Noether was a most remarkable woman! A. Einstein, H. Weyl, N. Wiener considered her the best female mathematician ever. Her bad luck was that she was a German Jew at the wrong time.  
I feel Ms. Noether deserves a link.

Peter230655:
Is eq(240) not also called a problem of the calculus of variation, with eq(241) as the solution?
If my memory is right, worth mentioning this?

Peter230655:
For someone like me, with virtually no prior exposure to Lagrange' method, this chapter is very concise.
Would it make sense to refer to some textbook, where this is explained in more detail?
I the lectures about Kane's method, the textbook by Prof. Kane was the reference, and it would be nice to give a similar reference here, I think.

wwolfie:
> Under Example: _3D body in outer space_, you give the code on how to calculate it. In previous chapters, the results were also given, I think, it would be nice to give them here, too.
> 
> I guess, the results would also explain the sentence right below the code:
> 
> ...but the terms, particularly the terms involving Missing argument for \dot $\dot \dot q_r$ are mangled.

Showing the results is nice, especially in this case. They can be a bit lengthy in general, so I will do things like checking the symbols used, and checking linearity. To get some useful (but brief) output.

wwolfie:
> Under the heading **Constrained equations of motion** you write: _This is not possible for non-holonomic constraint (by definition), and..._ I guess, if the non-holonomic constraints are really 'holonomic constraints in disguise' (in a previous chapter it was shown how to check this) then, by definition, this is not a non-holonomic system?

Yes, holonomic means the constraint _can_ be expressed as function of position only.

wwolfie:
> > For example, here we is used frequently, as in we have learned earlier...
> 
> We have a new co-author here, so there will be some style differences.
> 
> Edit: but we should work on consistency.

Yes, we should try and be as consistent as possible. I have removed almost all occurences of "we" as a start.

Peter230655:
I knew we have a coauthor for this chapter. I did not want to make it an issue, it was merely an observation.

wwolfie:
> For someone like me, with virtually no prior exposure to Lagrange' method, this chapter is very concise. Would it make sense to refer to some textbook, where this is explained in more detail? I the lectures about Kane's method, the textbook by Prof. Kane was the reference, and it would be nice to give a similar reference here, I think.

Yes, I think that is a great idea. The course that thought me (a previous course in Delft), used Lanczos' Variational Principles of Mechanics as source sometimes. That book is a bit old-fashioned, and written more for physicists instead of engineers. However, it has a good deal of depth,  with (to me) a clear writing style. Most importantly,  it sticks to relatively basic maths compared to some alternatives, which rely heave on differential geometry.

I added that one, but am still looking for other options

lagrange.rst
```diff
@@ -0,0 +1,631 @@
+============================================
+Equations of Motion with the Lagrange Method
+============================================
+
+.. note::
+
+   You can download this example as a Python script:
+   :jupyter-download-script:`generalized-forces` or Jupyter Notebook:
```
moorepants, line 8:
```suggestion
   :jupyter-download-script:`lagrange` or Jupyter Notebook:
```

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+   :jupyter-download-notebook:`generalized-forces`.
```
moorepants, line 9:
```suggestion
   :jupyter-download-notebook:`lagrange`.
```

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+Inertial forces from kinetic energy
+===================================
+
+When learning about Kane's method, we learned that the negated generalized inertial
+forces equal the applied forces, see :ref:`Unconstrained Equations of Motion`.
+A large part of Kane's method of deriving the equations of motions for a 
+system is involved with finding the generalized inertial forces.
+
+As an alternative, the following equation also get the generalized inertial forces of a
+system, now by starting from the `kinetic energy TODO`_ :math:`T(\dot{\bar{q}}, \bar{q})`
```
moorepants, line 65: It is possible to write this as T(u, q, t) and still retain the ability have kinematical differential equations other than q'=u. I've taught it like this before, but it will require `xreplace()` calls in places.

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+
+- Derive the Lagrangian for a system of particles and rigid bodies
+- Use the Euler-Lagrange equation to derive equations of motions given a Lagrangian
+- Use the method of Lagrange multipliers to add constraints to the equations of motions
+
+Introduction
+============
+
+This book has already discussed three methods to derive the equations
+of motion of multibody systems: Newton-Euler, Kane's method and the TMT
```
moorepants, line 50: Link to these places in the text where we introduce these other methods.

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+method. This chapter will add a third: the Lagrange method, originally 
+developed by Joseph-Louis Lagrange.
```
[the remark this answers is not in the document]
Peter230655: Should it not be: this chapter adds a **forth**...?

lagrange.rst
moorepants: I see you have u in the equation, so it should be consistent either way.

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+expressed as function of the generalized coordinates :math:`\bar{q}`, and 
+their time derivatives.
+
+.. _`kinetic energy TODO`: https://en.wikipedia.org/wiki/Work_in_process
+
+
+.. math::
+  :label: eq-lagrange-inertial
+
+   -\bar{F}^*_r = \frac{\mathrm{d}}{\mathrm{d}t}\left(\frac{\partial T}{\partial u_r}
```
moorepants, line 75:
```suggestion
   -\bar{F}^*_r = \frac{\mathrm{d}}{\mathrm{d}t}\left(\frac{\partial T}{\partial u}
```

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+        \right) - \frac{\partial T}{\partial q_r}
+
+.. warning:: Note the two minus signs in the above equation
+
+.. note::
+
+   In Kane's method, we are free to choose generalized speeds :math:`\bar{u}` that differ from
+   the time derivatives of the generalized coordinates, that is :math:`\dot{\bar{q}}`. This
+   freedom is not readily there when using the Lagrange method. Therefore we will consistently use
+   :math:`\dot{\bar{q}}` in this chapter.
```
moorepants, line 85: Then change the notation above. You can use things other than q'=u with Lagrange, so I'd just say by convention u=q' is assumed for Lagrange.

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+
+The generalized inertial forces computed in this manner are the same as when following
+the TMT method. This can be shown by carefully matching terms in both formulations, as
+is done for a system of point-masses in [Vallery2020]_.
+
+Example: 3D body in outer space
```
moorepants, line 91: If this is the same as https://moorepants.github.io/learn-multibody-dynamics/eom.html#body-fixed-newton-euler-equations, then maybe reference that and compare to it. I was able to display the equations there (if you use a body fixed expression)

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+
+.. jupyter-execute::
+
+   # Setting up reference frames
```
moorepants, line 95: I try to make separate jupyter cells in general and then move the comments to actual markdown text.

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+   psi,theta, phi, x, y, z = me.dynamicsymbols('psi theta phi x y z')
+   N = me.ReferenceFrame('N')
+   B = me.ReferenceFrame('B')
+   B.orient_body_fixed(N, (psi, theta, phi), 'zxy')
```
moorepants, line 99: This would make it different than my prior example, as these will be ignorable coordinates for a torque free rigid body. That's probably why your equations are long.

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+
+   # Mass and inertia
+   m, Ixx, Iyy, Izz = sm.symbols('M, I_{xx}, I_{yy}, I_{zz}')
+   I_B = me.inertia(B, Ixx, Iyy, Izz)
+
+   # Kinematics and kinetic energy
+
+   omega_B = B.ang_vel_in(N)
+   r_com = x*N.x + y*N.y + z*N.z
+   v_com = r_com.dt(N)
```
moorepants, line 109: I have common notation for the posistions and velocities. Consistency would be good.

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+   T = omega_B.dot(I_B.dot(omega_B))/2 + m*v_com.dot(v_com)/2
```
moorepants, line 110: I need to choose a variable name for kinetic, potential, and total energy. T and V are common for the first two, but I also use T (or bar{T}) for torque, which may be confusing. What do you think of E_k, E_p, and E_t?

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+
+   # Euler-Lagrange equation
+
+   t = me.dynamicsymbols._t
+   q = sm.Matrix([psi, theta, phi, x, y, z])
+   qdot = q.diff(t)
+   qddot = qdot.diff(t)
+   p = sm.Matrix([T]).jacobian(qdot).transpose()
+   g = -sm.Matrix([T]).jacobian(q).transpose()
```
moorepants, line 119: This is a bit obtuse for the first statement on forming these.

You need to show some math for this too I think.

g is always used for gravity, so that may not be a good variable name.

need to define "p" mathematically

Why are transposes needed here?

moorepants: The first time I show Lagrange I usually do the equation more expliclty and one by one:

```
T.diff(q1d).diff(t) - V.diff(q1)
```

This matches the equation form more closely.

Then after than I write the equations with Jacobian mathematics and then Jacobian code.

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+
+
+This gives the equations of motion, but the terms, particularly the terms
+involving :math:`\ddot{q}_r` are mangled. It is common to extract the system
+mass matrix and velocity forces vector like so:
+
+.. jupyter-execute::
+
+   mass_matrix = left_hand_side.jacobian(qddot)
+   dynamic_bias = left_hand_side - mass_matrix*qddot
```
moorepants, line 130: If you use trigsimp on these you may be able to display them and it may not take too long to compute.

The word "dynamic_bias" has not been introduced and there is no math to connect these lines.

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+   \bar{F}_r = -\frac{\partial V}{\partial q_r}
+
+.. warning:: Note the minus sign in the above equation.
+
+.. _`conservative forces`: https://en.wikipedia.org/wiki/Conservative_force
+.. _`potential energy`: https://en.wikipedia.org/wiki/Potential_energy
+
+Some examples of conservative forces are:
+
+* linearized gravity on the surface of the earth, with potential :math:`m g h(\bar{q})`,
```
moorepants, line 152: I usually say "uniform gravitational field", I'm not sure what "linearized" means here

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+* gravity from Newton's universal gravitation, with potential :math:`-G \frac{m_1m_2}{r(\bar{q})}`,
+* a linear spring, with potential :math:`\frac{1}{2}k(l(\bar{q}) - l_0)`.
```
moorepants, line 154: missing squared?

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+
+For conservative forces, it is often convenient to derive the applied forces via 
+the potential energy.
+
+
+The Lagrange-method
```
moorepants, line 160:
```suggestion
The Lagrange Method
```

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+===================
+
+Both the equation for computing the inertial forces from the kinetic energy, and 
+the equation for computing the applied forces from a potential energy have a term
+in them with the partial derivative with respect to the generalized coordinate. 
+Furtermore, the potential energy does not depend on the generalized speeds. 
+Therefore, we can derive the resulting (inertial and conservative applied) forces
+in one go, by combining the two equations.
+
+Step 1. Compute the so called Lagrangian :math:`L`, the difference between the
```
moorepants, line 170: wikipedia link to lagrangian

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+
+.. math::
+   :label: eq-euler-lagrange
+
+   \frac{\mathrm{d}}{\mathrm{d}t}\left(\frac{\partial L}{\partial u_r}
+       \right) - \frac{\partial L}{\partial q_r} = \bar{F}_r,
+    
+while being careful to include a force either in the applied forces 
+:math:`\bar{F}_r`, or in the potential energy :math:`V`, but never
+in both.
```
moorepants, line 189: This is an important note. Should we introduce some notation that breaks Fr into potential related forces and non-conservative ones?

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+   L = T - V
+
+Step 2. Use the Euler-Lagrange equations (the name for the equation 
+:ref:`eq-lagrange-inertial`) to find the equations of motion:
+
+.. math::
+   :label: eq-euler-lagrange
+
+   \frac{\mathrm{d}}{\mathrm{d}t}\left(\frac{\partial L}{\partial u_r}
+       \right) - \frac{\partial L}{\partial q_r} = \bar{F}_r,
```
moorepants, line 185: I like how you use Fr here, nice way to avoid having to show them how to arrive at generalized forces (the method I used was based on work and virtual displacements). Cool!

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+Because further examples include multiple bodies, we introduce two convenience functions to
+simplify the code for computing the kinetic energy:
+
+.. jupyter-execute::
+
+   def squarednorm(a):
+       return a.dot(a)
+
+   def quadraticform(I, v):
+       return v.dot(I.dot(v))
```
moorepants, line 203: These don't save much typing. I suggest not doing this. I also don't do things like this elsewhere, hoping that the sympy notation is reasonably close to the math notation.

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+
+   T = m/2*(squarednorm(Ao.vel(N)) + squarednorm(Bo.vel(N)) + squarednorm(Q.vel(N))) + 1/2*(
+       quadraticform(I_A_Ao, A.ang_vel_in(N)) + quadraticform(I_B_Bo, A.ang_vel_in(N))
+   )
+   V = m*g*(Ao.pos_from(O).dot(-N.x) + Bo.pos_from(O).dot(-N.x)) + kt/2*(q1**2) + kt/2*q2**2 + kl/2*q3**2
+
+   L = sm.Matrix([T - V])
+   lhs = L.jacobian(qdot).diff(t) - L.jacobian(q)
+   M = lhs.transpose().jacobian(qddot)
+   G = lhs.transpose() - M*qddot
```
moorepants, line 251: These should be M_d and g_d that I've defined before, right? If so let's stick with that notation.

moorepants: I recommend following the same style of replacing all qddot with zero to get G, then you should end up with some simpler equations that you could probably display.

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+is not possible for non-holonomic constraint (by definition), and requires to solve often difficult
+non-linear equations when considering holonomic constraints. This method of elimination is therefore
+not useful within the Lagrange method.
+
+Instead, we generalize the approach in :ref:`Exposing Noncontributing Forces`. We will first ommit the
+constraint, and add a constraint force, for which we can specify the direction, but not the magnitude. 
+The (second) time derivative of the constraint equation is then added to the equations found with the
+Euler-Lagrange equation.
+
+For a particle of mass :math:`m` and position :math:`\bar{r}_{P/O} = q_1 \hat{n}_x + q_2 \hat{n}_y + q_3\hat{n}_z` on a
```
moorepants, line 278:
```suggestion
For a particle of mass :math:`m` and position :math:`\bar{r}^{P/O} = q_1 \hat{n}_x + q_2 \hat{n}_y + q_3\hat{n}_z` on a 
```

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+    \end{array}
+
+This can be put in matrix-form, by extracting the unknown acceleration and force magnitude;
+
+.. math::
+    \begin{bmatrix} m & 0 & 0 &-1 \\ 0 & m & 0 & 1 \\ 0 & 0 & m & 0 \\ 1 & -1 & 0 & 0\end{bmatrix}
+    \begin{bmatrix} \ddot{q}_1 \\ \ddot{q}_2 \\ \ddot{q}_3 \\ F \end{bmatrix} = \begin{bmatrix} 0 \\ 0 \\ -mg \\ 0\end{bmatrix}
+
+
+It can be tricky to find the direction of the constraint force from the geometric of the system directly.
```
moorepants, line 298:
```suggestion
It can be difficult to find the direction of the constraint force from the geometric of the system directly.
```

to avoid saying "trick" twice

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+There is a trick, called the method of the Lagrange multupliers, to quickly find the correct generalized
+forces associated with the constraint forces. 
+
+Given a constraint in the general form
+
+.. math::
+
+    \sum_r a_r(\bar{q}) \dot{q}_r = 0
```
moorepants, line 306: i may have some notation for this: eq 207 https://moorepants.github.io/learn-multibody-dynamics/holonomic-eom.html maybe we need some way to connect this with that.

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+
+We find the generalized force as:
+
+.. math::
+
+    F_r = \lambda a_r(\bar{q})
+
+Here :math:`\lambda` is a variable encoding the magnitude of the constraint force. It is
+called  the Lagrange multiplier. The same :math:`\lambda`` is used for each :math:`r`, that is,
```
moorepants, line 315:
```suggestion
called  the Lagrange multiplier. The same :math:`\lambda` is used for each :math:`r`, that is, 
```

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+so, we make use of a useful fact, which is true for all systems and constraints:
+
+.. jupyter-execute::
+
+    diff_constraint.jacobian(qddot) - A
+
+This allows us to create our equations in a block matrix form:
+
+.. math::
+        \begin{bmatrix} M & A^T \\ A & 0\end{bmatrix}\begin{bmatrix}\ddot{\bar{q}} \\ \lambda \end{bmatrix} =
```
moorepants, line 404: let's use M_d and g_d also use 'mathbf{M}_d` for matrices

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+independent generalized velocities only. This can give rise to simplified equations, additional insight, and
+numerically more efficient simulation.
+
+Furthermore, the connection from Kane's method to vector mechanics, that is, Newton's law's, is clearer, which
+can provide additional insight, and make it easier to encorporate non-conservative forces such as friction.
+
+On the other hand, the Lagrange-method results in a set of equations with well understood structures and properties.
+These structures and properties are not studied further in these materials, other than the following "learn more" section.
+With further study, these aspects can make it easier to generalize results across multiple systems, for example
+when designing control algorithms.
```
moorepants, line 448: Writing the energy expressions only requires you to write magnitudes of velocities, which avoid having to work out accelerations explicitly. Writing energy expressions may be simpler for some problems than all the steps in Kane's method.

With Kane's you don't need Lagrange multipliers.

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+
+.. note::
+
+   In Kane's method, we are free to choose generalized speeds :math:`\bar{u}` that differ from
+   the time derivatives of the generalized coordinates, that is :math:`\dot{\bar{q}}`. This
+   freedom is not readily there when using the Lagrange method. Therefore we will consistently use
+   :math:`\dot{\bar{q}}` in this chapter.
+
+The generalized inertial forces computed in this manner are the same as when following
+the TMT method. This can be shown by carefully matching terms in both formulations, as
```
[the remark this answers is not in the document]
wwolfie: Yes, it would be the same. I wanted to refer out for a "proof" (by example) that they are the same, and just new this one. 

Have added that the Kane's method leads to the same result.

lagrange.rst
wwolfie: Added a reference, and explained the similarities and differences. Also see next comment.

lagrange.rst
wwolfie: Yes, the trouble I run into is that I want to just use u = \dot{q}. 

I don't know a clean method (so not the manual substitution used [ for example here](https://phys.libretexts.org/Bookshelves/Classical_Mechanics/Variational_Principles_in_Classical_Mechanics_(Cline)/13%3A_Rigid-body_Rotation/13.18%3A_Lagrange_equations_of_motion_for_rigid-body_rotation). To get back to the nice Euler-Equations

lagrange.rst
wwolfie: Changed them, and checked for others. Probably I missed some. Will keep doing iterations.

lagrange.rst
wwolfie: Yes, most people doing Lagrangian mechanics don't use external torques I guess. Happy with using E_subscripts. Can you confirm when you stick to those?

lagrange.rst
wwolfie: Did the first one. Will do a single example of the second one.

lagrange.rst
wwolfie: The slight advantage of the method I use is that in principle we can check if lhs is indeed linear in the qddots, by checking what symbols exists.

That said, for consistency, and easy understanding, I have changed it.

lagrange.rst
wwolfie: I'm showing the mass matrix, with some simplification. That looks ok. Now showing the dynamic bias, but it remains a bit lengthy

wwolfie: I also replaced dynamic-bias with g_d, with explanation that there is a slight difference. Might need to have a different variables, just as with F_r.

moorepants:
> For example, here we is used frequently, as in we have learned earlier...
> This was never used in earlier chapters.

Actually, I use it on 284 lines!

```
$ grep -r "we " --include "*rst" | wc -l
284
```

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+   Ao = me.Point("A_c")
+   Ao.set_pos(O, -0.5*l*A.z)
+   Bo = me.Point("B_c")
+   Bo.set_pos(Ao, -0.5*l*A.z - 0.5*l*B.z)
+   Co = me.Point("C_c")
+   Co.set_pos(Bo, -0.5*l*B.z -0.5*l*C.z)
+
+The next step is again to form the Lagrangian and find the equations of motion. As the system has no further constraints, 
+the Lagrange multiplier method is not needed. The actuator torques are added to the right hand side of the equation, in
+the same way as active forces are added to Kane's equations. Here the torques are represented by the variables :math:`T_b`
```
moorepants, line 506:
```suggestion
the same way as generalized active forces are added to Kane's equations. Here the torques are represented by the variables :math:`T_b`
```

lagrange.rst
```diff
@@ -0,0 +1,631 @@
...
+    \frac{\text{d}}{\text{d}t}\frac{\partial L}{\partial \dot{q}} - \frac{\partial L}{\partial q}= 0,
+
+which we recognize as the Euler-Lagrange equations.
+
+This means that the laws of nature governing rigid body motions result in motions that minimize the integral of the
+Lagrangian.  This is called Hamilton's principle. It turns out that many physical laws take such a form of minimizing
+the value of a function. One example is Fermat's principle, which states that light takes the path of minimum time.
+
+The optimization point-of-view of the Lagrange method also gives an interpretation for the Lagrange multipliers. They
+are the same as the Lagrange multipliers used in optimization.
```
moorepants, line 625: I didn't check these two learn more sections in detail, but my main comment is that we don't really have these extensive extra sections in the other chapters. Right now I can tell the students that anything from the chapters we covered can be on the exam. But adding this material makes that easy statement a bit more complex. So my initial reaction is not to include these extensive learn more sections. There are lots of "appendix" materials that I'd like to add (things that don't fit in the classes learning objectives) and we need a way to add them, but we haven't designed that yet. Thoughts?

moorepants: One more thought: these could be turned in homework problems, then we don't have anything that is "extra".

Peter230655: Most of the math books I read have a section _Notes_ or similar at the end of each chapter. There generally it is stated in which direction the material of the chapter can be extended, maybe some historical notes, etc. Not required reading, but interesting.

moorepants: Yes, we've added some to the chapters, but only some basic bullet lists to link out to things. These are a bit elaborate with examples and all, so I'm reluctant to include in the chapter. I've tried to make the chapters mostly the minimal set of things needed for the weekly lessons in the course so that students are too distracted with extra info when trying to solve the homeworks.

wwolfie: Yes, I agree with removing them. I'll soon push an updated version.

There, I added a very brief paragraph on "generalized momentum", and added knowing it to the learning goals. This paragraph still mentions Noether, and refers to ignorable coordinates (which were defined in the nonholonomic constraint chapter).

Also moved the reference to variational principles to the introduction.

I plan to use the "falling cat" in the homework.

moorepants:
@WouterWolfslag, how would you like to be listed as a co-author? Here are some options:

1. We list you and me as co-authors on the main page.
2. I list you as a contributing author.
3. We make the main work sort of authorless and then make a primary author for each chapter


Thoughts? I'm open for all of them and want to give you any recognition you feel you deserve.

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+.. math::
+        \begin{bmatrix} \mathbf{M}_d & \mathbf{A}^T \\ \mathbf{A} & 0\end{bmatrix}\begin{bmatrix}\ddot{\bar{q}} \\ \bar{\lambda} \end{bmatrix} = 
+        \begin{bmatrix} \bar{F}_r - \bar{g}_d \\ - \frac{\partial \mathbf{A}\dot{\bar{q}}}{\partial \bar{q}}\dot{\bar{q}} \end{bmatrix},
+
+where :math:`A` is the jacobian of the constraints, as used above,  :math:`\bar{g}` is the dynamic bias, and 
+the last term on the right hand side can be computed as;
+
+.. jupyter-execute::
+
+    constraint_bias = diff_constraint - diff_constraint.jacobian(qdd)*qdd
```
TJStienstra, line 462: Was just reading quickly through the chapter to get some inspiration for the multiple choice questions. However this line seems a bit weird/unexpected to me. If I translate the equation before literally I would expect:
```
- (diff_constraint.jacobian(qdd) * qd).jacobian(q) * qd
```

wwolfie: Yes, your way works as well. I think there are three (easy) ways to get the bias:

1. Just compute what it should be (like what you do)
2. Set qddot = 0 from the derivative of the constraint (_I will change it to this, because this method is used in similar situations elsewhere in the book_)
3. Substract the parts involving qddot from the second derivative of the constraint (as I did here, I normally do this because it allows me an easy check to see if my assumption on linearity in qddot is correct).

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+Introduction
+============
+
+This book has already discussed two methods to derive the equations
+of motion of multibody systems: Newton-Euler and Kane's method. This
+chapter will add a third: the `Lagrange method`_, originally 
+developed by Joseph-Louis Lagrange. These materials focus on Engineering
+applications for multi-body systems, and therefore build the Lagrange method around
+the terms found earlier in Kane's equations. In other textbooks, the Lagrange method
+is often derived from the `Variational principles`, such as virtual work or the principle
```
moorepants, line 56:
```suggestion
is often derived from the `Variational principles`_, such as virtual work or the principle
```

lagrange.rst
moorepants: Switch T to K, since I forced your hand :)

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+   :math:`\bar{u} = \dot{\bar{q}}` is assumed when using the Lagrange method. Therefore, throughout
+   this chapter :math:`\dot{\bar{q}}` is used.
+
+The generalized inertial forces computed in this manner are the same as when following
+Kane's method, or the TMT method used in the next chapter. This can be shown by carefully
+matching terms in these formulations, as is done for a a system of point-masses in [Vallery2020]_.
+
+Example: freely moving 3D body
+
+This example is largely the same as the example in `Body Fixed Newton-Euler Equations`_. A key difference
```
moorepants, line 102:
```suggestion
This example is largely the same as the example in :ref:`Body Fixed Newton-Euler Equations`. A key difference
```

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+   gd = Fs.xreplace(qdd_zerod)
+   Md.simplify()
+   gd.simplify()
+   Md, gd
+
+
+Conservative Forces
+===================
+
+Some applied forces, known as conservative forces `conservative forces`_, can
```
moorepants, line 170:
```suggestion
Some applied forces, known as `conservative forces`_, can
```

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+Therefore, the resulting (inertial and conservative applied) forces can be derived
+in one go, by combining the two equations.
+
+Step 1. Compute the so called `Lagrangian`_ :math:`L`, the difference between the 
+kinetic energy and potential energy:
+
+.. math::
+   :label: eq-lagrangian
+
+   L = T - V
```
moorepants, line 210:
```suggestion
   L = K - V
```

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+
+.. _`Lagrangian`: https://en.wikipedia.org/wiki/Lagrangian
+
+Step 2. Use the Euler-Lagrange equations (the name for the equation 
+:ref:`eq-lagrange-inertial`) to find the equations of motion:
```
moorepants, line 215:
```suggestion
:math:numref:`eq-lagrange-inertial`) to find the equations of motion:
```

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+
+   \frac{\mathrm{d}}{\mathrm{d}t}\left(\frac{\partial L}{\partial u_r}
+       \right) - \frac{\partial L}{\partial q_r} = \bar{F}_r,
+    
+while being careful to include a force either in the applied forces 
+:math:`\bar{F}_r`, or in the potential energy :math:`V`, but never
+in both.
+
+
+Example: Double pendulum with springs and sliding pointmass
```
moorepants, line 228: header?

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+   L = sm.Matrix([T - V])
+   left_hand_side = L.jacobian(qd).diff(t) - L.jacobian(q)
+
+   qdd_zerod = {qddr: 0 for qddr in qdd}
+   Md = left_hand_side.jacobian(qdd)
+   gd = left_hand_side.xreplace(qdd_zerod)
+
+   me.find_dynamicsymbols(Md), me.find_dynamicsymbols(gd)
+
+The mass matrix :math:`M` only depends on :math:`\bar{q}`, and :math:`\bar{g}` depends
```
moorepants, line 282:
```suggestion
The mass matrix :math:`\mathbf{M}_d` only depends on :math:`\bar{q}`, and :math:`\bar{g}_d` depends
```

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+coordinates, as defined in `Equations of Motion with Nonholonomic Constraints`_. 
+
+Constrained equations of motion
+===============================
+
+When using Kane's method, constraints are handled by dividing the generalized speeds into two sets:
+the dependent and independent generalized speeds. Then, the dependent generalized speeds are eliminated 
+by solving the (time derivative of the) constraint equation.
+
+In the Lagrange method, the generalized speeds should always match the generalized coordinates.
```
moorepants, line 307: Maybe also say that the number of dynamical differential equations always equals the number of kinematical differential equations in Lagrange's method. Or something about the number of final equations. Kane's method take a system with N coordinates and give you N kinematical differential equations, p dynamical differential equations, and M holonomic algebraic constraint equations where N - M = n and n - m = p.

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+Therefore, to handle constraints, the generalized coordinates should be likewise eliminated. This
+is not possible for non-holonomic constraint (by definition), and requires to solve often difficult
```
moorepants, line 309:
```suggestion
is not possible for holonomic constraint (by definition), and requires to solve often difficult
```

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+non-linear equations when considering holonomic constraints. This method of elimination is therefore
+not useful within the Lagrange method.
+
+Instead, constraints are handled using a generalized version of the approach in 
+:ref:`Exposing Noncontributing Forces`. First the constraints are omitted. Then a constraint force is added,
```
moorepants, line 314:
```suggestion
:ref:`Exposing Noncontributing Forces`. First the constraints are omitted. Then a constraint force is added,
```

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+with a known direction, but unknown magnitude. Finally, the (second) time derivative of the constraint 
+equation is then appended to the equations found with the Euler-Lagrange equation.
+
+For example, consider a particle of mass :math:`m` and position 
+:math:`\bar{r}_{P/O} = q_1 \hat{n}_x + q_2 \hat{n}_y + q_3\hat{n}_z`
```
moorepants, line 319:
```suggestion
:math:`\bar{r}^{P/O} = q_1 \hat{n}_x + q_2 \hat{n}_y + q_3\hat{n}_z`
```

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+
+.. math::
+    \begin{array}{r}
+    m\ddot{q}_1= \phantom{-}F\\
+    m\ddot{q}_2= -F\\
+    m\ddot{q}_3 + mg = \phantom{-}0\\
+    \ddot{q}_1 - \ddot{q}_2\!\! = \phantom{-}0 
+    \end{array}
+
+This can be put in matrix-form, by extracting the unknown acceleration and force magnitude;
```
moorepants, line 335:
```suggestion
This can be put in matrix-form, by extracting the unknown acceleration and force magnitude:
```

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+.. math::
+    \begin{bmatrix} m & 0 & 0 &-1 \\ 0 & m & 0 & 1 \\ 0 & 0 & m & 0 \\ 1 & -1 & 0 & 0\end{bmatrix}
+    \begin{bmatrix} \ddot{q}_1 \\ \ddot{q}_2 \\ \ddot{q}_3 \\ F \end{bmatrix} = \begin{bmatrix} 0 \\ 0 \\ -mg \\ 0\end{bmatrix}
+
+
+It can be challenging to find the direction of the constraint force from the geometry of the system directly.
+There is a trick, called the method of the Lagrange multupliers, to quickly find the correct generalized
+forces associated with the constraint forces. 
+
+Given a constraint in the general form
```
moorepants, line 346:
```suggestion
Given a motion constraint (time derivative of a configuration constraint or a nonholonomic constraint) in the general form
```

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+    \sum_r a_r(\bar{q}) \dot{q}_r = 0
+
+The generalized force is found as:
+
+.. math::
+
+    F_r = \lambda a_r(\bar{q})
+
+Here :math:`\lambda` is a variable encoding the magnitude of the constraint force. It is
+called  the Lagrange multiplier. The same :math:`\lambda` is used for each :math:`r`, that is,
```
moorepants, line 359: link to https://en.wikipedia.org/wiki/Lagrange_multiplier here?

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+For example, consider the pointmass to be constrained to move in a bowl :math:`x^2 + y^2 + z^2 -1 = 0`.
+Taking the time derivative gives: :math:`a_1 = 2q_1` :math:`a_2 = 2q_2` and :math:`a_3 = 2q_3`.
+This results in generalized reaction forces :math:`F_1 = 2\lambda q_1`, :math:`F_2 = 2\lambda q_2` and :math:`F_3 = 2\lambda q_3`.
+
+
+Example: turning the freely floating body discussed earlier into a rolling sphere.**
+
+The non-slip condition of the rolling sphere is split into three constraints: the velocity of
+the contact point (:math:`G`) is zero in the :math:`\hat{n}_x`, the :math:`\hat{n}_y` and  the :math:`\hat{n}_z`
+direction. These constraints are enforced by contact forces in their respective directions.
```
moorepants, line 377: We should be explicit about this being a configuration constraint or a nonholonomic constraint. You say the velocity is constrained.

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+    \bar{n}_x\cdot ({}^N\bar{v}^C + {}^N\bar{\omega}^B \times (-r\hat{n}_z)) = 0 \\
+    \bar{n}_y\cdot ({}^N\bar{v}^C + {}^N\bar{\omega}^B \times (-r\hat{n}_z)) = 0 \\
+    \bar{n}_z\cdot ({}^N\bar{v}^C + {}^N\bar{\omega}^B \times (-r\hat{n}_z)) = 0 \\
+    \end{array}
+
+These can be used to derive the constraint force and the additional equations using the Lagrange-multiplier
+method as shown below. Note that here only the first time derivative of the constraint equation is used, 
+again because the second time derivatives of the generalized coordinates appear.
+
+.. container:: invisible
```
moorepants, line 394: do a dropdown admonition here in case they need to see the setup

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+        qdd = qd.diff(t)
+
+        L = sm.Matrix([T])
+        left_hand_side = L.jacobian(qd).diff(t) - L.jacobian(q)
+
+        qdd_zerod = {qddr: 0 for qddr in qdd}
+        Md = left_hand_side.jacobian(qdd)
+        gd = left_hand_side.xreplace(qdd_zerod)
+
+To make these free floating body a rolling wheel, three constraints are needed: the
```
moorepants, line 429:
```suggestion
To make this free floating body a rolling wheel, three constraints are needed: the
```

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+velocity of the contact point should be zero in :math:`\hat{n}_x`, :math:`\hat{n}_y`
+and :math:`\hat{n}_x` direction.
+
+.. jupyter-execute::
+
+    lambda1, lambda2, lambda3 = me.dynamicsymbols('lambda1, lambda2, lambda3') 
+    constraint = (v_com + B.ang_vel_in(N).cross(-N.z)).to_matrix(N)
+    A = constraint.jacobian(qd)
```
moorepants, line 437: You use lowercase math a in the text and then a cap A here. Can you make this code `A` match a `$A$` in latex math somewhere?

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+    diff_constraint = constraint.diff(t)
+    constraint
```
moorepants, line 439:
```suggestion
    sm.trigsimp(constraint)
```

this works nicely for a compact result

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+This equality is true for all constraints, as can easily be shown by taking the time
+derivative of the constraint equation, using the chain rule.
+
+The combined equations can now be written in a block matrix form:
+
+.. math::
+        \begin{bmatrix} \mathbf{M}_d & \mathbf{A}^T \\ \mathbf{A} & 0\end{bmatrix}\begin{bmatrix}\ddot{\bar{q}} \\ \bar{\lambda} \end{bmatrix} = 
+        \begin{bmatrix} \bar{F}_r - \bar{g}_d \\ - \frac{\partial \mathbf{A}\dot{\bar{q}}}{\partial \bar{q}}\dot{\bar{q}} \end{bmatrix},
+
+where :math:`A` is the jacobian of the constraints, as used above,  :math:`\bar{g}` is the dynamic bias, and
```
moorepants, line 457:
```suggestion
where :math:`\mathbf{A}` is the jacobian of the constraints, as used above,  :math:`\bar{g}` is the dynamic bias, and 
```

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+The method of the Lagrange multiplier can of course also be used within Kane's method. However,
+it results in a larger system of equations, which is why the elimination approach is often
+preferred there. An exception being scenarios where the constraint force itself is a useful output,
+for instance to check no-slip conditions in case of limited friction.
+
+
+Lagrange's vs Kane's
+====================
+
+The is book has now thought to alternatives to the Newton-Euler method: Kane's method and Lagrange's method.
```
moorepants, line 476:
```suggestion
The is book has now presented two alternatives to the Newton-Euler method: Kane's method and Lagrange's method. 
```

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+
+
+Lagrange's vs Kane's
+====================
+
+The is book has now thought to alternatives to the Newton-Euler method: Kane's method and Lagrange's method. 
+This raises the questions: when should each alternative method be used?
+
+For constrained systems, Kane's method has the advantage that the equations of motion are given for a set of
+independent generalized velocities only. This can give rise to simplified equations, additional insight, and
```
moorepants, line 480:
```suggestion
independent generalized speeds only. This can give rise to simplified equations, additional insight, and
```

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+Furthermore, the connection from Kane's method to vector mechanics, that is, Newton's law's, is clearer, which
+can provide additional insight, and make it easier to encorporate non-conservative forces such as friction.
+
+On the other hand, the Lagrange method only requires energies velocities as input, for which only the velocities
+of the bodies are needed. This is can be simpler to derive than the accelerations which are needed for Kane's
+method.
+
+Furthermore, the Lagrange method results in a set of equations with well understood structures and properties.
+These structures and properties are not studied further in these materials. A starting point for further study
+is `Noether's theorem`, which extends the idea of ignorable coordinates to find conserved quantities like
```
moorepants, line 493:
```suggestion
is `Noether's theorem`_, which extends the idea of ignorable coordinates to find conserved quantities like
```

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+
+    diff_constraint.jacobian(qdd) - A
+
+This equality is true for all constraints, as can easily be shown by taking the time
+derivative of the constraint equation, using the chain rule.
+
+The combined equations can now be written in a block matrix form:
+
+.. math::
+        \begin{bmatrix} \mathbf{M}_d & \mathbf{A}^T \\ \mathbf{A} & 0\end{bmatrix}\begin{bmatrix}\ddot{\bar{q}} \\ \bar{\lambda} \end{bmatrix} =
```
moorepants, line 454: define A mathematically before you use it in the code

moorepants:
@WouterWolfslag this looks solid now. I gave a bunch of minor fixes as comments, but nothing major is needed.

Let me know what I can do to help finish this, if anything. Nice work!

lagrange.rst
moorepants: > Substract the parts involving qddot from the second derivative of the constraint (as I did here, I normally do this because it allows me an easy check to see if my assumption on linearity in qddot is correct).

I think this may not always result in the symbolic cancellations that you hope for without simplification. That's why I do the qdd = 0 method in other places.

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+    diff_constraint.jacobian(qdd) - A
+
+This equality is true for all constraints, as can easily be shown by taking the time
+derivative of the constraint equation, using the chain rule.
+
+The combined equations can now be written in a block matrix form:
+
+.. math::
+        \begin{bmatrix} \mathbf{M}_d & \mathbf{A}^T \\ \mathbf{A} & 0\end{bmatrix}\begin{bmatrix}\ddot{\bar{q}} \\ \bar{\lambda} \end{bmatrix} = 
+        \begin{bmatrix} \bar{F}_r - \bar{g}_d \\ - \frac{\partial \mathbf{A}\dot{\bar{q}}}{\partial \bar{q}}\dot{\bar{q}} \end{bmatrix},
```
moorepants, line 455: is the constraint bias term missing parentheses around the first q'?

moorepants: I have Jacobian notation in the notation page. Would it be helpful to just the Jacobian notation in this chapter? Then maybe you don't even need A or you can define A as the Jacobian notation of `\dot{f}_h`.

lagrange.rst
wwolfie: Rewrote the sentence to clarify. It really should be non-holonomic in this part though

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+.. math::
+
+    P = \sum_r F_r\dot{q}_r = \sum \lambda a_r(\bar{q})\dot{q}_r  = \lambda \sum a_r(\bar{q})\dot{q}_r = \lambda \cdot 0
+
+For example, consider the pointmass to be constrained to move in a bowl :math:`x^2 + y^2 + z^2 -1 = 0`.
+Taking the time derivative gives: :math:`a_1 = 2q_1` :math:`a_2 = 2q_2` and :math:`a_3 = 2q_3`.
+This results in generalized reaction forces :math:`F_1 = 2\lambda q_1`, :math:`F_2 = 2\lambda q_2` and :math:`F_3 = 2\lambda q_3`.
+
+
+Example: turning the freely floating body discussed earlier into a rolling sphere.**
```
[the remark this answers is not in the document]
wwolfie: ** Was a leftover from making it bold. Now a sub-header

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+
+.. jupyter-execute::
+
+    constraint_bias = diff_constraint - diff_constraint.jacobian(qdd)*qdd
+
+We call the block matrix called the extended mass matrix, and the vector on the right hand side the extended dynamic bias. 
+
+
+The method of the Lagrange multiplier can of course also be used within Kane's method. However,
+it results in a larger system of equations, which is why the elimination approach is often
```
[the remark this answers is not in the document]
wwolfie: Yes, I used 'larger' as in 'larger number of equations' but that is ambiguous.

lagrange.rst
wwolfie: 1. No, it is a matrix multiplication. I write it inside the partial, because that way everything stays one or two dimensional. Otherwise we'd have to define how to deal with derivatives of a matrix valued function. 
2. To handle earlier comment, I defined a new $\boldsymbol{M}_{hn}$, similar to what is done for holonomic constraints. I use it here now as well.

lagrange.rst
```diff
@@ -0,0 +1,692 @@
...
+preferred there. An exception being scenarios where the constraint force itself is a useful output,
+for instance to check no-slip conditions in case of limited friction.
```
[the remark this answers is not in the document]
wwolfie: Added equation counts throughout the chapter, and fixed the use of "length" for number of equations.

wwolfie:
Thanks for all the feedback! 

I think I have fixed everything, although I will double check for consistency and accuracy of the formulations again later. If I have time, I'll add some figures as well.

moorepants:
>  If I have time, I'll add some figures as well.

I can sketch some out if you need help on that. Let me know.

eom.rst
```diff
@@ -335,7 +335,7 @@ velocities and angular velocities of :math:`A_o`, :math:`B_o`, :math:`A`, and
    B = me.ReferenceFrame('B')
 
    A.orient_axis(N, q1, N.z)
-   B.orient_axis(A, q2, A.x)
+   B.orient_axis(A, q2, A.z)
```
moorepants, line 338: Is this an error?

wwolfie: Actually no, I have now convinced myself the original is correct. The perspective just messes with my (and E's) head.

wwolfie: Yes, I agree, have pushed a commit reverting to the original: https://github.com/moorepants/learn-multibody-dynamics/pull/156/commits/81c4397e2aebb70fb81f2422af5c4a62089d8329

lagrange.rst
```diff
@@ -0,0 +1,773 @@
...
+Then, set up the Lagrangian:
+
+.. jupyter-execute::
+
+   t = sm.symbols('t')
+   q = sm.Matrix([q1, q2, q3])
+   qd = q.diff(t)
+   qdd = qd.diff(t)
+
+   K = m/2*(Ao.vel(N).dot(Ao.vel(N)) + Bo.vel(N).dot(Bo.vel(N)) + Q.vel(N).dot(Q.vel(N))) + 1/2*(
```
moorepants, line 285:
```suggestion
   K = m/2*(Ao.vel(N).dot(Ao.vel(N)) + Bo.vel(N).dot(Bo.vel(N)) + Q.vel(N).dot(Q.vel(N))) + sm.S(1)/2*(
```

moorepants: Or move the divide by 2 at the end of the expression.

wwolfie: I think the divide by 2 suggestion is a bit simpler. Should we add a hint to the sympy chapter for this as well?

moorepants: I show it in the sympy chapter already (first warning on the page).

moorepants: sorry not the first one, but there

moorepants:
Congrats :confetti_ball: !

Peter230655:
In the first code box below the heading **Example: freely moving 3D body you write**:
I_B = me. inertia(B, Ixx, Iyy, Izz)
In previous lectures, the inertia was always formed with the tensors, with outer.
Worth mentioning that _me.inertia(..)_ does just this?

Peter230655:
Below the printout of the equations of motion you write:
_While these are correct equations of motion, the terms, particularly the terms involving q¨r are mangled_

I cannot find q¨r in these equations. Do you mean the sm.Matrix q defined earlier?

Peter230655:
Below Some examples of conservative forces are: you write:

_a linear spring, with potential V=1/2k(l(q¯)−l0)2._
In the Hunt_Crossley method used with the jumper in a previous lecture, the spring force is non-linear, but still conservative. Of course with a different potential.
Worth mentioning, that non-linear springs are also conservative?

Peter230655:
In the description of Fig 53 you write:
_When the generalized coordinates are all zero, the two rods are perpendicular to each other_
I thought, rod B rotates around a_x.
If I understand this correctly, then the two rods are always perpendicular to each other, no matter what values the generalized coordinates take?

Peter230655:
In the code box below **Then, set up the Lagrangian** you write:
_t = sm.symbols('t')_
I was told by JM, that this is a 'dangerous' way to define t, one should write
t = me.dynamicsymbols._t ?

Peter230655:
Below the heading **Constraint equations of motion** you write:
_The method of elimination is therefore not useful within the Lagrange method._
As I understood Kane's method, there are the same problems with holonomic constraints, difficult non-linear equations to be solved.

For my understanding:
Is the idea of converting q1 - q2 = 0 into q1¨−q2¨=0 similar to what was done earlier, converting a holonomic constraint into speed constraints (here acceleration constraints)?

PR #157, opened 2023-05-11
Update holonomic equation of motion chapter

moorepants:


Peter230655:
Below eq(206) you write:
_We can formulate the equations of motion by transforming the holonomic constraints into a function of generalized speeds._
I assume, this means forming $\frac{d}{dt}eq(205) = 0$ and solve the resulting linear system for $\bar \dot q$.

Peter230655:
It took me a while to understand the dimension of eq(209).
Likely so, because in sympy mechanics the u_r are explicit in the state vector, and then 'removed' with the velocity_constraints keyword, equivalent to eq(208).
If what I said is correct, then I understood 209, else I have to think some more.

Peter230655:
Under the heading **Simulate without constraint enforcement** you write:
_As presented, Eqs. (209) only contain the constraints that the velocity and acceleration of point P4 must be zero,...
_It is clear, that the veocity must be zero, this is eq(208), but I do not see the _explicit_ condition for the acceleration.
Im I overlooking something obvious?
Of course, if v(t) = 0 for all t, then d/dt(v(t)) = 0

Peter230655:
Somewhere further down, you write:
_SciPy’s fsolve() function is capable of finding the roots for sets of nonlinear equations, given a good guess._
A while back I tried this four link pendulum in 3D. I found it impossible to get a good initial guess by looking at the system.
I finally used scipy's minimize(configuration constraint), and used the result as my initial guess. This worked reliably in my example. (The initial guess for minimize(..) seems less crucial)

PR #160, opened 2023-05-31
Added figure of bowl constraint forces and changed x, y, z to q2, q2, q3.

moorepants:


moorepants:
@WouterWolfslag , this is to merge into your branch

PR #167, opened 2024-02-12
Minor wording updates to the introduction for 2024.

moorepants:


Peter230655:
We have discussed before, you disagreed, and I accept, still I want to say it:

..... so that you can use them as a tool to answer scientific questions and
solve engineering problems.
smypy,physics.mechanics is the GREAT tool to apply your lecture to state and solve these problems! Should it not be mentioned somewhere?

moorepants:
I can't toot my own horn that hard :)

Peter230655:
😄 😄 
then drop GREAT, but mention it somewhere. It is a fantastic tool to set up multi body system - I have played around with it enough to know.

moorepants:
I may make an extra chapter that shows the mechanics module, but I don't want to distract the students at the beginning of the class.

Peter230655:
This would be good and important, I think: even my non-realistic examples often had count_ops numbers of the force vector exceeding 50,000. No way to do this without sympy.mechanics.
So, to apply what they learned in your lecture to real world problems (as you say, they will learn) they NEED this tool.

Peter230655:
If you want me to do so, I could try to write such a chapter (I could only do it on Jupyter notes, not on sphinx, or whatever you use) and you take, change it, correct it or discard it completely.

moorepants:
I'm not completely sure I want to add one about this. You have to remember that the primary learning objectives for the course this book goes with is to learn the formulation of multibody dynamics, i.e. they need to write every equation themselves. The course isn't about using a tool, I more want them to be able to create the tool. If I add this chapter about mechanics and they see it immediately in the class, then they may skip learning how to form the equations of motion themselves.

Peter230655:
This makes sense!
Would it be possible to not have this chapter visible until the lecture is (almost) over? 
I should not push, but I find it such a great tool to apply what they learned to real world problems.

moorepants:
There is actually a lot there: https://docs.sympy.org/dev/modules/physics/mechanics/index.html#guide-to-mechanics but it could use some improved organization and updated writing.

Peter230655:
This IS a lot, much more than I would know! So, for my 'concern' it should be sufficient if you point this out at the end of your lecture, even verbally. My concern is that the students learn all basic theory - and then miss this tool

Peter230655:
I think, my concern is unwarranted: I am nearly 69 years old. The stundents are young. They are interested in multi body dynamics, this is why they take the course. Surely, they will find sympy.physics.mechanics easily! :-)

PR #168, opened 2024-02-12
Minor updates for the jupyter-python chapter for 2024.

moorepants:


Peter230655:
_output of the code. You may also have a text cell that contains static content.
Cell type can be changed using a Jupyter drop-down menu_.

Would it make sense to mention latex here? Took me a long time to figure out the use of latex - but maybe young students know about it anyway.

moorepants:
> Would it make sense to mention latex here? Took me a long time to figure out the use of latex - but maybe young students know about it anyway.

I added a note, but they should follow the "learn more" section to learn all the jupyter goodies.

PR #171, opened 2024-02-25
Updated wording in some areas of the angular kinematics chapter for 2024.

moorepants:


Peter230655:
In this:
https://moorepants.github.io/htmlpreview/?https://github.com/moorepants/learn-multibody-dynamics/blob/preview-pr-171/index.html
when I click [Kane 1985] , above ea (54) it does not get me to the references as before. It gets me back to the top of the chapter.

Peter230655:
$Aω¯B:=(Adb^ydt⋅b^z)b^x+(Adb^zdt⋅b^x)b^y+(Adb^xdt⋅b^y)b^z.$

Peter230655:
Of course, eq (54) is a definition, and not be be argued with.
But, for example, the term:
$\left(\dfrac{{}^Âd\hat b_y }{dt} \circ \hat b_z \right) \hat b_x$

moorepants:
Some things may not work in the preview.

moorepants:
> Upon second thought: the abstract definition (54) gives all the more intuitive results as is shown further down in the lecture.

Where further down do you think it shows this?

I think I'll open a separate issue for this, as I won't be able to address it immediately.

PR #175, opened 2024-03-11
Added an explanation of sign conventions in figure dimensions.

### LINKED ISSUE #50: Create an explanation of the sign conventions in drawings
Students are confused about whether to add a + or - to rotations based on my drawings. I need to show some drawings and explain the convention I use.

moorepants:
Fixes #50


Peter230655:
In the picture where you explain the rotation would it make sense to indicate the direction of $\hat n_z$ (pointing at the observer) or will it overload the drawing?

moorepants:
I currently do not have a way to indicate arrows that point directly out of or into the page. I don't have a notation for that (I don't think).

moorepants:
I do show one in figure 11 https://moorepants.github.io/learn-multibody-dynamics/vectors.html but I don't think I ever use that in my figures.

Peter230655:
At least for me, the most important rule in these matters in the 'right hand rule', which you show in one of your pictures.
I use it all the time to decide in which direction a moment vector turns the body.

moorepants:
I explain the right hand rule in the text in more than one place. The reason I'm adding this section is that the students are continually confused about the sign conventions used in my figures (in their homework problems). They have most issues with the angular dimensions and how to translate that to a positive or negative right handed rotation.

moorepants:
In Figure 19, I use the -q notation (which I'd like to get rid of):
image

moorepants:
Fig 28 doesn't use the dimension leaders:
image

moorepants:
Fig 31 uses no arrows on linear dimensions:
image

Peter230655:
Fig 28 looks very clear to me.

moorepants:
Yes, that may be, but it should be consistent, especially if I introduce this new notation explanation.

Peter230655:
Makes sense!

moorepants:
I've finalized this and am going to allow adding a -q or -F as sometimes it makes drawing the figures more convenient to match with positive right-handed rotations.

Peter230655:
I guess, a good idea:
- how would you want to draw the lamp of fig 19? It would look like it would fall over immediately
- in addition, it may help students to realize that angles are not automatically positive - as one might assume from high school math.
- maybe even shows that the right hand rule is important.

moorepants:
> how would you want to draw the lamp of fig 19? It would look like it would fall over immediately

I decided to "allow" the use of `-q` to explain that the drawn configuration is in fact the result of a negative right handed rotation. So I won't draw the lamp in a only positive rotation configuration.

Peter230655:
This is what I meant by 'using' the right hand rule.
When I first started with sympy.mechanics, I had to really be careful about this - and made many mistakes.

moorepants:
I added a couple more sentences based on @Peter230655 and Kariens' feedback. I'm going to merge this, as I need to share with the students to clear up some questions. Thanks for the help.

Peter230655:
In the orientation chapter you have eq(11). Would it make sense to add this 'clockwise circle ($\hat n_x \rightarrow \hat n_y \rightarrow \hat n_z \rightarrow \n_x)$ as help? I mean similar to what they give with the quaternion units i, j, k?

moorepants:
I have that figure in the vector chapter.

moorepants:
Or, a similar figure.

Peter230655:
You mean this one?
https://upload.wikimedia.org/wikipedia/commons/thumb/3/34/Right-hand_grip_rule.svg/240px-Right-hand_grip_rule.svg.png

PR #176, opened 2024-03-14
Correct sign error in rolling without slip explanation.

moorepants:


Peter230655:
I believe you had.

$\bar_x \times \bar n_y = -\bar n_x$

Maybe $..\times (-r \bar n_y)$ would look a bit nicer than without brackets?

Peter230655:
Applying the motion constraint and knowing that $Nv¯G=0$
 gives us this scalar constraint equation

PR #198, opened 2025-02-11
Week 2025 01

moorepants:


moorepants:
This should have been:

```python
coef_mat, rhs = sm.linear_eq_to_matrix(exprs, *unknowns)
```

the unknowns have to be expanded.

PR #228, opened 2026-04-17
Better explain the TMT calculation for inertia forces

### LINKED ISSUE #217: Update the TMT chapter to calculate the C in Mq'' + C(q') = F using the Jacobian of the constraints
Right now I do the naive approach which is to just take the time derivative and then substitute zero for the qdots. But you can find C with something like J.T@J to get the velocity cross terms.

moorepants:
Fixes #217

Shows more carefully how to calculate the inerital forces that are linear in the velocities (Coriolis, gyroscopic, centripital). It is worth noting that Euler's equation is:

dH/dt = Iw' + w x (Iw)

and is equivalent to:

dH/dt = I'w + Iw'

but when grouping linear and rotational equation in one single big matrix equation you can't write the cross product form because it only applies to a 3D vector. You can probably write this succintly with the Clifford Algebra based dynamics approaches.

It still seems that papers leave out the I'w term (but that might because they are only about 2D systems). This confused me several times.

### DIFF
--- a/tmt.rst
+++ b/tmt.rst
@@ -246,18 +246,6 @@ where:
    \bar{g} = \frac{d\mathbf{M}\bar{v}}{dt}\bigg\rvert_{\dot{\bar{u}}=\bar{0}} =
    \left(\dot{\mathbf{M}} \mathbf{T} + \mathbf{M} \dot{\mathbf{T}}\right) \bar{u}
 
-The equations of motion then take this form:
-
-.. math::
-
-   \mathbf{M}_d\dot{\bar{u}} + \bar{g}_d =
-   -\mathbf{T}^T \mathbf{M} \mathbf{T} \dot{\bar{u}} +
-   \mathbf{T}^T\left(\bar{F} - \bar{g}\right)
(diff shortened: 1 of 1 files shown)


moorepants:
@Peter230655 @phdejong you both may be interested in this. If either of you are willing to review my explanation and math on this one, that is appreciated.

tmt.rst
```diff
@@ -211,26 +236,24 @@ with:
...
 .. math::
 
-   \bar{0} =
    \mathbf{M}_d\dot{\bar{u}} + \bar{g}_d =
    -\mathbf{T}^T \mathbf{M} \mathbf{T} \dot{\bar{u}} +
    \mathbf{T}^T\left(\bar{F} - \bar{g}\right)
+   = \bar{0}
 
-These equations are equivalent to Kane's Equations.
+These equations are equivalent to Kane's Equations and Lagrange's dynamical
```
moorepants:
```suggestion
These equations are equivalent to Kane's and Lagrange's dynamical
```

PR #232, opened 2026-04-24
Show how to use the An matrix in the Nonholonomic EoM chapter

moorepants:
- [x] Figure out how to show both methods: 1) substitute ur from the get go and 2) apply An.
- [x] Adapt the code to support the two methods.
- [x] Settle on new notation.

### DIFF
--- a/nonholonomic-eom.rst
+++ b/nonholonomic-eom.rst
@@ -92,9 +92,10 @@ originally developed in Sec. :ref:`Snakeboard`:
 .. math::
    :label: eq-dep-speeds-repeat
 
-   \bar{u}_r = \mathbf{A}_n \bar{u}_s + \bar{b}_n\\
+   \bar{u}_r = \mathbf{A}_n \bar{u}_s + \bar{b}_n \in \mathbb{R}^m
 
-If you partition the unconstrained dynamical differential equations into:
+If you partition the holonomic (unconstrained) dynamical differential equations
+into:
 
 .. math::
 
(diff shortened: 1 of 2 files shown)


moorepants:
@Peter230655 if you are interested, I made significant modifications to this chapter and any read through with review and feedback is very welcome. I will teach it Monday May 11 in class and have to finalize by then. It is already merged so you can see it on the website.

Peter230655:
> @Peter230655 if you are interested, I made significant modifications to this chapter and any read through with review and feedback is very welcome. I will teach it Monday May 11 in class and have to finalize by then. It is already merged so you can see it on the website.

I will start right away.

Peter230655:
If I have any comments / questions should I put them here or in an email to you?

moorepants:
Here is fine.

Peter230655:
Above eq (201):
....as long as $A_r$ is invertible.
If $A_r$ is singular, likely there was a modeling error, too many dependent variables, I assume.
(I had such singularity issues when a played around with nonlinear velocity constraints)
Maybe a footnote adressing this?

Below eq (205)
_...differentiation index 1_
Do your students know what this is? I only learned it since I joined a lecture on DAEs
Maybe a footnote adressing this?

Above eq205)
_Then using (180) and (187) the p nonholonomic....._
Maybe you could add a few steps to explain how to get from (180), (187) to (205), after all, (205) is important.

moorepants:
Ar can be singular from unfortunate choice of coordinates and generalized speeds for a specific configuration (as well as what you mention).

I removed the mention of the differentiation index.

I think 205 arises directly from 180 and 187, there aren't really any steps between.

Peter230655:
> Ar can be singular from unfortunate choice of coordinates and generalized speeds for a specific configuration (as well as what you mention).

In my cases mentioned, I could always (as I recall) fix it by a better choice of coordinates. 
I still feel that in a footnote you could mention  reasons why $A_r$ may be singular, after all you indicate it may be singular.
> 
> I removed the mention of the differentiation index.

> 
> I think 205 arises directly from 180 and 187, there aren't really any steps between.

It was not obvious to me today. I will check again tomorrow and see if I will see it then.

Peter230655:
About a page below _Approach 2_   (there is no equation anywhere close to point to the location) you write

 Calculate $A_n$ from the nonholonomic constraints.

You give no explanation, how you do it. I assume you get it from eq(121), wq(122) (?)
Maybe a line of explanation would be useful.

moorepants:
I can refer back to the equations in the motion.rst chapter.

moorepants:
>  If A r is singular, likely there was a modeling error, too many dependent variables, I assume.

I will leave this out, as I don't have it formulated yet how to write exactly all reasons this can be invertible.

